PR #1, opened 2017-09-25
fix division left arg rule to ensure consistency with right arg rule

jrevels:
This should fix https://github.com/JuliaDiff/ForwardDiff.jl/issues/264.

Thanks for pointing this out, @simonbyrne


simonbyrne:
Will this fix it?

jrevels:
Ah, it doesn't, good catch. I had put parens in the wrong place when testing this locally 😛

PR #4, opened 2017-10-30
Allow diff rule function to be in an expression or QuoteNode

jdlangs:
This fixes the current broken build and keeps the package compatible with `v0.6` and `v0.7` after the change in expression parsing that puts `Symbol`s in `QuoteNode`s instead of `Expr` objects (JuliaLang/julia#23885).

I also changed a few references to left-hand side (LHS) and right-hand side (RHS) which seemed backwards. Let me know if anything is missing or erroneous.

Thanks!

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -12,7 +12,7 @@ struct DiffRule{M,f} end
 Define a new differentiation rule for the function `M.f` and the given arguments, which should
 be treated as bindings to Julia expressions.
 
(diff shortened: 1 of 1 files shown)


jrevels:
Thanks for this!

> I also changed a few references to left-hand side (LHS) and right-hand side (RHS) which seemed backwards

Woah, I must've been having a pretty weird day when I wrote that code! Thanks for fixing those.

PR #9, opened 2017-12-20
Derivatives for `Base.transpose` and `Base.\`

wesselb:
Like the title says, this PR adds derivatives for `Base.transpose` and `Base.\`.


jrevels:
Thanks (and sorry for the late response)!

PR #11, opened 2018-03-03
Add max, min, and abs.

willtebbutt:
What the title says.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -60,7 +60,7 @@
 @define_diffrule Base.lgamma(x)               = :(  digamma($x)                        )
 @define_diffrule Base.Math.JuliaLibm.log1p(x) = :(  inv($x + 1)                        )
 @define_diffrule Base.transpose(x)            = :(  1                                  )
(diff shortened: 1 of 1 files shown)


src/rules.jl
```diff
@@ -60,6 +60,7 @@
 @define_diffrule Base.lgamma(x)               = :(  digamma($x)                        )
 @define_diffrule Base.Math.JuliaLibm.log1p(x) = :(  inv($x + 1)                        )
 @define_diffrule Base.transpose(x)            = :(  1                                  )
+@define_diffrule Base.abs(x)                  = :( $x > 0 ? one($x) : -one($x)         )
```
jrevels: Should we use `signbit` instead here (since, e.g. `!(-0.0 < 0.0)`)? At least, this is what ForwardDiff's manual definition currently uses, though IIRC that decision was made fairly arbitrarily at the time...

willtebbutt: That makes sense. Will make the change.

willtebbutt:
@jrevels requested changes made. Is there anything else you would like to happen prior to merging?

jrevels:
Nope, sorry for the delay. Thanks!

Issue #14, opened 2018-06-18
`log1p` getting overwritten


ExpandingMan:
There seems to be an extra method from another library [here](https://github.com/JuliaDiff/DiffRules.jl/blob/master/src/rules.jl#L66).  At least on 0.7 this results in the methods for `log1p` getting overwritten.  Would it be ok to either delete this line or, if it's needed for 0.6, move it to within the conditional?

jrevels:
Yes, we should just wrap it in a conditional e.g. 

```julia
if is_version_that_doesnt_need_to_reference_libm(VERSION)
    @define_diffrule Base.log1p(x) = ...
else
    @define_diffrule Base.Math.JuliaLibm.log1p(x) = ...
end
```

I don't know what that predicate is off the top of my head, but it seems like just `Base.log1p` works for v0.7.

jrevels:
closed by #15

PR #15, opened 2018-06-19
fixed method overwrite of `log1p`

ExpandingMan:
This is a possible fix for #14.

I don't know if the version being used here is correct.  I *do* know that on the first 0.7 release candidate, making this definition was giving a method overwrite error.  I'd be happy to change if anyone knows what the proper condition for this definition is, or where I might be able to find that.


jrevels:
LGTM. We can always change this later if any unforeseen issues pop up. The next DiffRules release that gets tagged after v0.7 is released will probably drop v0.6, so no worries.

Thanks!

PR #18, opened 2018-06-28
Qualify special functions with module name.

### LINKED ISSUE #17: clarify @define_diffrule namespaces
I think that eg
```julia
@define_diffrule SpecialFunctions.lgamma(x) = :( digamma($x) )
```
should be
```julia
@define_diffrule SpecialFunctions.lgamma(x) = :( SpecialFunctions.digamma($x) )
```
since the RHS is not necessarily evaluated in the module of the LHS. Is this correct?

Just please clarify, I would be happy to make a PR.

(cf https://github.com/JuliaStats/StatsFuns.jl/issues/49)

tpapp:
Fixes #17.

Lines were broken because they became too long.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -61,8 +61,10 @@ if VERSION < v"0.7-"
     @define_diffrule Base.lgamma(x)           = :(  digamma($x)                        )
     @define_diffrule Base.Math.JuliaLibm.log1p(x) = :(  inv($x + 1)                    )
 else
-    @define_diffrule SpecialFunctions.gamma(x) = :(  digamma($x) * gamma($x)           )
-    @define_diffrule SpecialFunctions.lgamma(x) = :(  digamma($x)                      )
+    @define_diffrule SpecialFunctions.gamma(x) =
+        :(  SpecialFunctions.digamma($x) * SpecialFunctions.gamma($x)  )
+    @define_diffrule SpecialFunctions.lgamma(x) =
+        :(  SpecialFunctions.digamma($x)  )
 end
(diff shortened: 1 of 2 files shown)


tpapp:
Someone knowledgeable **please** review the line

```julia
@define_diffrule NaNMath.lgamma(x) = :(  SpecialFunctions.digamma($x)               )
```

because I am not sure about the corner cases.

jrevels:
I assume that that's fine, but I'll ask @mlubin since his packages - e.g. JuMP - are the main users of these definitions.

mlubin:
Actually JuMP doesn't currently use DiffRules, although we'd like to. But anyway is the issue that digamma is missing from NaNMath?

tpapp:
The issue is that 
```julia
ForwardDiff.derivative(NaNMath.lgamma, x)
```
may be nonsensical for `x` where `NaNMath.lgamma` differs from `SpecialFunctions.lgamma`, since it uses `SpecialFunctioncs.digamma`. But it may just work if the analytic extension matches. I don't know enough about the gamma function to answer this.

In any case, unless there are objections, it would be great to merge this PR. It fixes some bugs, and I don't think it makes anything worse.

mlubin:
I don't know enough about lgamma or digamma either, so no objection here.

andreasnoack:
Would be great to get this one in since it is holding back downstream packages.

jrevels:
> Actually JuMP doesn't currently use DiffRules, although we'd like to.

JuMP uses ForwardDiff, which uses DiffRules. But anyway, this seems fine. Thanks again @tpapp!

PR #21, opened 2018-08-07
dep fix: Random.seed! instead of Random.srand

StefanKarpinski:


StefanKarpinski:
@KristofferC, @keno, can one of you merge this?

Keno:
I don't have direct access. @jrevels would though.

Issue #10, opened 2018-02-20
Supporting Linear Algebraic Primitives


jrevels:
(continued from https://github.com/invenia/Nabla.jl/issues/81, cc @willtebbutt)

> This means that we would need a slightly more general interface for linear algebra, and would certainly require different forward- and reverse- mode expressions, than is currently provided by DiffRules.

Agreed, DiffRules only properly handles scalar kernels now. To support linear algebra, we need to add a notion of tensor/scalar, allowing in-place methods, marking adjoint variables, etc. to DiffRules.

> Regarding the where possible statement above, there are certainly operations (such as the Cholesky factorisation) for which symbolic expressions might be a bit unweildy (see here). Now that I think about this a bit more, I'm not sure whether these larger implementations are going to be problematic or not.
>
> Anyway, my point is that given symbolic expressions for the linear algebra operations I agree that it's reasonable to hope that compiler optimisations can eliminate redundant code when compiling custom kernel implementations, and that this is a significantly better idea than hand-coding lots of optimisations. (The issue I linked in my previous comment is a good example of this. I would definitely prefer to be able to just forget about this). However, I would contend that you simply can't handle linear algebra properly without a number of hand-coded symbolic expressions for the forward- and reverse-mode sensitivities because they aren't written in Julia. If at some point in the future we have native Julia implementation of (for example) LAPACK, then it would be a really good idea to try and produce an AD tool which is able to produce reasonably-well optimised kernels for each operation. To the best of my knowledge, we shouldn't expect this to happen any time soon (and almost certainly never for BLAS), so a symbolic version of the current implementation of DiffLinearAlgebra will be necessary for Capstan to be able to differentiate arbitrary Julia code even reasonably efficiently.

[...]

PR #22, opened 2018-08-11
WIP: AD stuff

willtebbutt:
Work in progress PR to extent DiffRules to provide useful AD-related functionality. The scope of the additions include:
- A collection of forward- and reverse-mode sensitivity implementations. These will comprise the existing scalar implementations and linear algebra optimisations for both modes.
- New syntax for specifying sensitivities.

Initial commit:
- Refactor `@define_diffrule` macro to improve readability using MacroTools.
- Define initial interface for adding forward sensitivity implementations.
- Construct forward rules from existing diff rules.

ToDo:
- [x] reverse-mode rules
- [x] `wrt` API. See #10. Requires a more structured representation of the rules than I currently provide.
- [ ] Add linear algebra sensitivities from [DiffLinearAlgebra.jl](https://github.com/invenia/DiffLinearAlgebra.jl)

Unknowns:
- specialisations for binary forward rules for when a sensitivity is provided only wrt one of the arguments. This requires a bit more structure in the sensitivity definitions. Is this necessary though? Could we just assume that zeros will be constant-propagated and rely on downstream packages to handle this appropriately?


jrevels:
Sweet, I'll try to review this this week. Also interested in thoughts from @MikeInnes and @dfdx, if they have any.

> Do we want to bump the require to 1.0.0? Tests are failing on 0.6. Do we care?

Nah, we can bump to 1.0 whenever.

> The coverage appears to have plummeted, but I have no idea why... also not sure if we care.

Would be good to know why but not essential at the moemnt

ararslan:
What's the status of this PR? Is this change still relevant going forward in the current Julia AD landscape? If so, I can try to help get it cleaned up.

willtebbutt:
This is PR is basically dead given [ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl/) as far as I'm concerned. I'm happy for this to be closed.

PR #29, opened 2018-11-22
Fix the examples in the docstrings

wookay:
the documentation needs to be updated.
http://www.juliadiff.org/DiffRules.jl/latest/#DiffRules.diffrule


jrevels:
Thanks!

PR #31, opened 2019-02-13
Add rules for NaNMath.min and max

andreasnoack:
Is it correct that these are automatically tested?


jrevels:
Yup, the test just takes all applicable rules and tests them against finite differences.

PR #33, opened 2019-02-19
add hook for abs definition that downstream packages can overload

jrevels:
This is the kind of hack that is obsoleted by ChainRules' method overloading approach vs a symbolic table, but in the meantime, this will enable @dpsanders to define an `abs` derivative for Intervals


dpsanders:
Great, thanks!

PR #36, opened 2019-06-25
Avoid breaking when x is negative

### LINKED ISSUE #27: Cannot backprop `x^a` when `x` is negative
I meet an AD problem in Flux.Tracker. Basically one cannot backprop `x^a` when `x` is negative, no matter whether or not a is even or odd. 

To be more specific, this works 
```julia
Tracker.gradient((x) -> x^2, 2.0) # => (4.0 (tracked),)
```` 
but this gives domain error 
```
Tracker.gradient((x) -> x^2, -2.0)

ERROR: DomainError with log:
-2.0 will only return a complex result if called with a complex argument. Try -2.0(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(::Float64, ::Symbol) at ./math.jl:31
 [2] log(::Float64) at ./special/log.jl:285
 [3] _forward at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/scalar.jl:55 [inlined]
 [4] #track#1 at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/Tracker.jl:50 [inlined]
 [5] track at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/Tracker.jl:50 [inlined]
 [6] log at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/scalar.jl:57 [inlined]
 [7] #218 at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/scalar.jl:66 [inlined]
 [8] back_(::Flux.Tracker.Grads, ::Flux.Tracker.Call{getfield(Flux.Tracker, Symbol("##218#219")){Flux.Tracker.TrackedReal{Float64},Int64},Tuple{Flux.Tracker.Tracked{Float64},Nothing}}, ::Int64) at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/back.jl:103
 [9] back(::Flux.Tracker.Grads, ::Flux.Tracker.Tracked{Float64}, ::Int64) at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/back.jl:118
 [10] #6 at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/back.jl:131 [inlined]
 [11] (::getfield(Flux.Tracker, Symbol("##9#11")){getfield(Flux.Tracker, Symbol("##6#7")){Flux.Tracker.Params,Flux.Tracker.TrackedReal{Float64}}})(::Int64) at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/back.jl:140
 [12] gradient(::Function, ::Float64) at /Users/kai/.julia/packages/Flux/UHjNa/src/tracker/back.jl:152
 [13] top-level scope at none:0
```
. 

I'm posting this issue here becuase I think this is related to https://github.com/JuliaDiff/DiffRules.jl/blob/master/src/rules.jl#L83. When `x` is negative, `log(x)` will of course throw an error. Is this behaviour expected? 

BTW ForwardDiff doesn't seem to have this problem.

### PR DESCRIPTION
**Pull request by thebhatman**
This PR is regarding this issue : https://github.com/FluxML/Zygote.jl/issues/247. So basically, Zygote is trying to calculate the derivative with respect to both base and exponent, due to the DiffRule mentioned for literal_pow. But when x is negative, it breaks due to domain error on log. So I feel it's better to check for positive base, before calculating the derivative as `f(x) = a ^ x` is differentiable only when a > 0.

Closes #27

### CODE REVIEW

**Comment by MikeInnes** [comment]
I'm guessing this will at least need to coerce the `NaN` to the right type. (e.g. the type of log(x)).

There's also some question as to whether this should be `NaN` or `0`. `NaN` seems like it could be annoying in practice, and possibly not available for all floating point types (?).

**Comment by arnavs** [comment]
How would I go about using this PR for my own Zygote work? Running into the same issue. Is it enough to check out the right branch of DiffRules?

**Comment by MikeInnes** [comment]
We should probably just define this rule in Zygote itself, overriding the DiffRules one. A PR would be great if someone's up for it.

**Comment by arnavs** [comment]
@MikeInnes Sorry, missed your reply on this. Would it be literally the exact same line of code?

**Comment by MikeInnes** [comment]
More or less. You do need to adapt it to how Zygote defines adjoints, which is in the docs.

File: src/rules.jl [hunk 1]
```diff
@@ -84,7 +84,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.:*(x, y) = :( $y                 ), :(  $x                )
 @define_diffrule Base.:/(x, y) = :( inv($y)            ), :( -($x / $y / $y)    )
 @define_diffrule Base.:\(x, y) = :( -($y / $x / $x)    ), :( inv($x)            )
-@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :(  ($x^$y) * log($x) )
+@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :( $x > 0 ? ($x^$y) * log($x) : NaN )
```
mcabbott, line 87: Since this package depends on NaNMath.jl and
```
julia> NaNMath.log(2.0f0)
0.6931472f0

julia> NaNMath.log(-2.0f0)
NaN32
```
the obvious thing to do is
```suggestion
@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :(  ($x^$y) *  $(NaNMath.log)($x) )
```

mcabbott: Bad idea. It depends on the package only to fix a version (I presume) but does not load it. So this doesn't work.

src/rules.jl
```diff
@@ -79,7 +79,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.:*(x, y) = :( $y                 ), :(  $x                )
 @define_diffrule Base.:/(x, y) = :( one($x) / $y       ), :( -($x / $y / $y)    )
 @define_diffrule Base.:\(x, y) = :( -($y / $x / $x)    ), :( one($y) / ($x)     )
-@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :(  ($x^$y) * log($x) )
+@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :(  ($x^$y) * $(NaNMath.log)($x) )
```
mcabbott, line 82: Maybe like this:
```suggestion
@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :( ($x isa Real && $x>0) ? ($x^$y)*log($x) : Base.oftype(float($x), NaN) )
```

src/rules.jl
```diff
@@ -79,7 +79,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.:*(x, y) = :( $y                 ), :(  $x                )
 @define_diffrule Base.:/(x, y) = :( one($x) / $y       ), :( -($x / $y / $y)    )
 @define_diffrule Base.:\(x, y) = :( -($y / $x / $x)    ), :( one($y) / ($x)     )
-@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :(  ($x^$y) * log($x) )
+@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :( ($x isa Real && $x>0) ? ($x^$y)*log($x) : Base.oftype(float($x), NaN) )
```
mcabbott, line 82: I think I got the logic wrong. Complex x is fine, it's only real non-positive that must avoid log. 
```suggestion
@define_diffrule Base.:^(x, y) = :( $y * ($x^($y - 1)) ), :( ($x isa Real && $x<=0) ? Base.oftype(float($x), NaN) : ($x^$y)*log($x) )
```

PR #37, opened 2019-07-10
Added support for complex differentiable functions

fp4code:
When a complex function of a complex variable is differentiable, `@define_diffrule` can be replaced with  `@define_complex_diffrule`. Then both `diffrule` and `complex_diffrule` will exist.


antoine-levitt:
I think this should be merged. Then ForwardDiff can use it to fix this very bad issue:
```
using ForwardDiff
f(x) = exp(2.0+im*x)
@show ForwardDiff.derivative(f, 0.0)
@show (f(1e-8) - f(0))/1e-8
```
This bug is because this just runs `exp(::Complex)` with a Complex{Dual}, and that method has a `if imag(x) == 0 ...`. The PR seems reasonable to me but probably has to be updated. @fp4code are you still interested in this and can you do it?

PR #38, opened 2019-10-14
Update for 1.0 world

oxinabox:
This:
 - Updates CI Scripts
 - Adds a direct dependency of SpecialFunctions, so it can  caps SpecialFunctions to v0.7, then will make follow up PR to unbreak it.
 - Removes documentation because that needs to be done differently (will open issue to fix it)


oxinabox:
idk if adding a direct dependency on SpecialFunctions so can cap version  is the best way.
or if should just add both the new names and the old names to the list of rules

oxinabox:
huh, Ok
I forgot making this

PR #52, opened 2020-11-22
Move SpecialFunctions to test deps

YingboMa:
This PR moves Move SpecialFunctions to test deps since it is not loaded in the source code.


YingboMa:
Superseded by #53

PR #53, opened 2020-11-22
Remove dependencies

mcabbott:
This package doesn't actually load SpecialFunctions.jl nor NaNMath.jl, so they can be test dependencies.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,20 +1,18 @@
 name = "DiffRules"
 uuid = "b552c78f-8df3-52c6-915a-8e097449b14b"
-version = "1.0.2"
+version = "1.0.3"
 
 [deps]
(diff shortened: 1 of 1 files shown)


oxinabox (changes requested):
DiffRules depends on SpecialFunctions so that it can set compat.
Because it doesn't define rules for all versions of SpecialFunctions, and SpecialFunctions has removed some things it has rules for.
So it needs that direct dependency so Pkg can make sure you get the right one.
This especially matters as people on Julia 1.0 can't use the latest version of SpecialFunctions, and thus also can't use the latest version of DiffRules.
But Pkg can handle that just fine, as long as you tell it.


Too bad Pkg will delete comments from Project.toml
Maybe we should add one anyway and just make sure it doesn't get deleted when people commit changes.

PR #55, opened 2021-01-17
Fixes return-types for multiple rules

torfjelde:
## Motivation
Currently, almost all of the rules using `π` currently has a return-type that differs from the input, e.g. `SpecialFunctions.erfcx` uses `√π` and so `SpecialFunctions.erfcx(1f0)` results in `Float64` rather than `Float32`.

## Aim
This PR fixes these issue by replacing all usages of `π` with `oftype($x, π)`. This is a rather naive and quick-fix.

## Up-for-discussion
- It *might* be preferable to instead exploit Julia's constant propagation, e.g. do `√π` is replaced by `oftype($x, √π)` instead of `√oftype($x, π)`, but I don't know enough about LLVM to know if this makes a difference such simple computations, hence I just went with naive conversion of `π` directly.
- It would be ideal to have tests for this, but it's unclear to me if all unary rules are expected to have the same return-type as the input-type or not. It seems to me like this ought to be the case, but didn't add any yet as I'm not 100% certain.


andreasnoack:
Thanks for fixing these. As a matter of style, I prefer to avoid explicit uses of `typeof` when possible. Sometimes it's the only way to get the type right but often you can simply adjust the order of evaluation in a way such that the promotion is correct. E.g. `(π / 180) * cosd($x)` could be changed to `cosd($x) / 180 * π`.

src/rules.jl
```diff
@@ -100,15 +100,15 @@ end
...
+    :(  (sqrt(oftype($x, π)) / 2) * exp(SpecialFunctions.erfinv($x)^2)  )
+@define_diffrule SpecialFunctions.erfc(x)        = :( -(2 / sqrt(oftype($x, π))) * exp(-$x * $x)       )
 @define_diffrule SpecialFunctions.erfcinv(x)     =
-    :( -(sqrt(π) / 2) * exp(SpecialFunctions.erfcinv($x)^2)  )
-@define_diffrule SpecialFunctions.erfi(x)        = :(  (2 / sqrt(π)) * exp($x * $x)        )
+    :( -(sqrt(oftype($x, π)) / 2) * exp(SpecialFunctions.erfcinv($x)^2)  )
+@define_diffrule SpecialFunctions.erfi(x)        = :(  (2 / sqrt(oftype($x, π))) * exp($x * $x)        )
 @define_diffrule SpecialFunctions.erfcx(x)       =
-    :(  (2 * $x * SpecialFunctions.erfcx($x)) - (2 / sqrt(π))  )
+    :(  (2 * $x * SpecialFunctions.erfcx($x)) - (2 / sqrt(oftype($x, π)))  )
```
andreasnoack, line 111: I think this could fail for integer inputs (unless there is some promotion before the rules are hit). Generally, it's safer to use "output" variables for the promotion so it would be something like
```
y = 2 * $x * SpecialFunctions.erfcx($x))
return y - oftype(y, 2)/π
```

src/rules.jl
```diff
@@ -25,26 +25,26 @@
...
 @define_diffrule Base.cot(x)                  = :( -(1 + cot($x)^2)                    )
-@define_diffrule Base.sind(x)                 = :(  (π / 180) * cosd($x)               )
-@define_diffrule Base.cosd(x)                 = :( -(π / 180) * sind($x)               )
-@define_diffrule Base.tand(x)                 = :(  (π / 180) * (1 + tand($x)^2)       )
-@define_diffrule Base.secd(x)                 = :(  (π / 180) * secd($x) * tand($x)    )
-@define_diffrule Base.cscd(x)                 = :( -(π / 180) * cscd($x) * cotd($x)    )
-@define_diffrule Base.cotd(x)                 = :( -(π / 180) * (1 + cotd($x)^2)       )
-@define_diffrule Base.sinpi(x)                = :(  π * cospi($x)                      )
-@define_diffrule Base.cospi(x)                = :( -π * sinpi($x)                      )
+@define_diffrule Base.sind(x)                 = :(  (oftype($x, π) / 180) * cosd($x)               )
```
mcabbott, line 28: Should we be concerned what happens if `x` is a substantial expression? This might evaluate it twice, which could be avoided by something like
```
@gensym z
@define_diffrule Base.sind(x)  = :(  $z=$x;  (oftype($z, π) / 180) * cosd($z)   )
```
I see that a few existing rules do already violate this, e.g. with `exp(-$x * $x)`. 

It's possible that `@define_diffrule` could be extended to take care of this. It's also possible that every downstream use calls CSE which will clean this up anyway.

src/rules.jl
torfjelde: Sorry for the delay here; been a busy couple of months and so this was pushed quite far down my priority list.

But yeah, this is a good point. Is the best solution then to go through all of the rules and manually make these changes, or is there a better way?

src/rules.jl
torfjelde: I wasn't aware that `x` could potentially be an non-evaluated expression; if so, that's for sure an issue, but maybe a separate one from this PR?

torfjelde:
Made some updates. There are a couple of things that:
1. Currently not testing binary methods properly because it's a bit unclear to how the types of the input arguments is decide in these tests. Seems like you want to sometimes have the first argument be a integer? 
2. Added a `@testset` wrapper because I was a bit confused as to what threw which error. IMO the output is more understandable if something fails, but this can be removed if you want.

torfjelde:
Actually, would it be better if I just added the constants needed, e.g. copy-paste those we need from https://github.com/JuliaStats/StatsFuns.jl/blob/master/src/constants.jl ?

That would solve almost all the promotion-bugs:)

torfjelde:
An update here: am currently trying to convince people that it's a good idea to make a MathConstants.jl package so that we can avoid explicit conversions here, in SpecialFunctions.jl, etc.

torfjelde:
I added the necessary irrationals to simplify the type-promotion. If you're happy to depend on IrrationalConstants.jl instead, we can make use of irrationals defined there instead:)

torfjelde:
Now that packages have started using IrrationalConstants.jl, maybe this is ready for a proper consideration? @andreasnoack @mcabbott

src/constants.jl
```diff
@@ -0,0 +1,7 @@
+import Base: @irrational
+import IrrationalConstants: logtwo
+
+@irrational logten 2.302585092994046 log(big(10))
+
+@irrational twoinvsqrtπ 1.1283791670955126 2 / sqrt(big(π))
+@irrational halfsqrtπ   0.886226925452758 sqrt(big(π)) / 2
```
torfjelde, line 7: I don't _really_ need these. I've made a PR to IrrationalConstants.jl to include `logten`, once that's gone through we can remove this file completely.

devmotion: The PR was merged and released in IrrationalConstants 0.1.1

devmotion:
What's the status of this PR? Can you resolve the conflicts and remove the constants since they are part of IrrationalConstants 0.1.1 @torfjelde?

torfjelde:
I'll give it a go :+1:

torfjelde:
I'm about to fly so I don't have time to wait for the results, but locally a couple of tests are failing and I'm a bit uncertain why.
- `SpecialFunctions.beta` seems to fail. Unclear if it's just because the `finitediff` isn't accurate enough when using `Float32`.

src/rules.jl
```diff
@@ -58,10 +58,9 @@
 @define_diffrule Base.acsch(x)                = :( -inv(abs($x) * sqrt(1 + $x^2))      )
 @define_diffrule Base.acoth(x)                = :(  inv(1 - $x^2)                      )
 @define_diffrule Base.sinc(x)                 = :(  cosc($x)                           )
-@define_diffrule Base.deg2rad(x)              = :(  π / 180                            )
-@define_diffrule Base.mod2pi(x)               = :(  isinteger($x / 2pi) ? NaN : 1      )
-@define_diffrule Base.rad2deg(x)              = :(  180 / π                            )
-
+@define_diffrule Base.deg2rad(x)              = :(  oftype($x, π / 180)                )
+@define_diffrule Base.mod2pi(x)               = :(  isinteger($x / 2pi) ? oftype($x, NaN) : one($x) )
```
devmotion, line 62: Similarly here, this does not work for general `x`. Maybe use something like
```suggestion
@define_diffrule Base.mod2pi(x)               = :(  isinteger($x / $twoπ) ? oftype(float($x), NaN) : one(float($x)) )
```

src/rules.jl
```diff
@@ -58,10 +58,9 @@
...
+@define_diffrule Base.rad2deg(x)              = :(  oftype($x, 180 / π)                )
```
devmotion, line 63: Similarly here:
```suggestion
@define_diffrule Base.rad2deg(x)              = :(  rand2deg(one($x))  )
```

src/rules.jl
```diff
@@ -112,18 +111,24 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 
 # unary #
 #-------#
-
-@define_diffrule SpecialFunctions.erf(x)         = :(  (2 / sqrt(π)) * exp(-$x * $x)       )
+@define_diffrule SpecialFunctions.erf(x) = :( 2 * exp(-$x * $x) * $(DiffRules.invsqrtπ) )
```
devmotion, line 114:
```suggestion
@define_diffrule SpecialFunctions.erf(x) = :( 2 * exp(-$x^2) * $invsqrtπ )
```

src/rules.jl
```diff
@@ -5,46 +5,46 @@
...
+@define_diffrule Base.:+(x)                   = :(  one($x)                            )
+@define_diffrule Base.:-(x)                   = :( -one($x)                            )
 @define_diffrule Base.sqrt(x)                 = :(  inv(2 * sqrt($x))                  )
 @define_diffrule Base.cbrt(x)                 = :(  inv(3 * cbrt($x)^2)                )
 @define_diffrule Base.abs2(x)                 = :(  $x + $x                            )
 @define_diffrule Base.inv(x)                  = :( -abs2(inv($x))                      )
 @define_diffrule Base.log(x)                  = :(  inv($x)                            )
-@define_diffrule Base.log10(x)                = :(  inv($x) / log(10)                  )
-@define_diffrule Base.log2(x)                 = :(  inv($x) / log(2)                   )
+@define_diffrule Base.log10(x)                = :(  inv($x) / $(DiffRules.logten)      )
```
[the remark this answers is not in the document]
devmotion: Or alternatively, just use
```suggestion
@define_diffrule Base.log10(x)                = :(  inv($x) / DiffRules.logten  )
```
but this seems to make more assumptions (one has to load DiffRules explicitly and not only e.g. `using DiffRules: diffrules`).

src/rules.jl
```diff
@@ -112,18 +111,24 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
...
-
-@define_diffrule SpecialFunctions.erf(x)         = :(  (2 / sqrt(π)) * exp(-$x * $x)       )
+@define_diffrule SpecialFunctions.erf(x) = :( 2 * exp(-$x * $x) * $(DiffRules.invsqrtπ) )
+@define_diffrule SpecialFunctions.erfinv(x) = :( (exp(SpecialFunctions.erfinv($x)^2) * $(DiffRules.sqrtπ)) / 2 )
+@define_diffrule SpecialFunctions.erfc(x) = :( -(exp(-$x * $x) * $(DiffRules.invsqrtπ)) * 2 )
+@define_diffrule SpecialFunctions.erfcinv(x) = :( -(exp(SpecialFunctions.erfcinv($x)^2) * $(DiffRules.sqrtπ)) / 2 )
+@define_diffrule SpecialFunctions.erfi(x) = :( 2 * exp($x * $x) * $(DiffRules.invsqrtπ) )
+@define_diffrule SpecialFunctions.erfcx(x) = :( 2 * ($x * SpecialFunctions.erfcx($x) - $(DiffRules.invsqrtπ)) )
+
+@define_diffrule SpecialFunctions.erf(x)         = :(  2 * (DiffRules.invsqrtπ * exp(-$x * $x))       )
```
devmotion, line 121: It's a bit inconsistent that you use `DiffRules....` instead of `$DiffRules` here.

src/rules.jl
```diff
@@ -211,14 +216,14 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
...
-@define_diffrule NaNMath.tan(x)    = :(  1 + NaNMath.pow(NaNMath.tan($x), 2)        )
-@define_diffrule NaNMath.asin(x)   = :(  inv(NaNMath.sqrt(1 - NaNMath.pow($x, 2)))  )
-@define_diffrule NaNMath.acos(x)   = :(  -inv(NaNMath.sqrt(1 - NaNMath.pow($x, 2))) )
-@define_diffrule NaNMath.acosh(x)  = :(  inv(NaNMath.sqrt(NaNMath.pow($x, 2) - 1))  )
-@define_diffrule NaNMath.atanh(x)  = :(  inv(1 - NaNMath.pow($x, 2))                )
+@define_diffrule NaNMath.tan(x)    = :(  1 + NaNMath.pow(NaNMath.tan($x), oftype($x, 2))        )
+@define_diffrule NaNMath.asin(x)   = :(  inv(NaNMath.sqrt(1 - NaNMath.pow($x, oftype($x, 2))))  )
+@define_diffrule NaNMath.acos(x)   = :(  -inv(NaNMath.sqrt(1 - NaNMath.pow($x, oftype($x, 2)))) )
+@define_diffrule NaNMath.acosh(x)  = :(  inv(NaNMath.sqrt(NaNMath.pow($x, oftype($x, 2)) - 1))  )
+@define_diffrule NaNMath.atanh(x)  = :(  inv(1 - NaNMath.pow($x, oftype($x, 2)))                )
```
devmotion, line 223: Are you sure about these rules? I would assume the current rules are correct and fine - they already account for the type of `x`. In fact, it seems more complicated to work with exponents of type e.g. `Float64` instead of `Int` and e.g. in Base it would mean that optimizations for `x^2` are not applied.
```suggestion
@define_diffrule NaNMath.tan(x)    = :(  1 + NaNMath.pow(NaNMath.tan($x), 2)        )
@define_diffrule NaNMath.asin(x)   = :(  inv(NaNMath.sqrt(1 - NaNMath.pow($x, 2)))  )
@define_diffrule NaNMath.acos(x)   = :(  -inv(NaNMath.sqrt(1 - NaNMath.pow($x, 2))) )
@define_diffrule NaNMath.acosh(x)  = :(  inv(NaNMath.sqrt(NaNMath.pow($x, 2) - 1))  )
@define_diffrule NaNMath.atanh(x)  = :(  inv(1 - NaNMath.pow($x, 2))                )
```

src/rules.jl
```diff
@@ -211,14 +216,14 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
...
 @define_diffrule NaNMath.log(x)    = :(  inv($x)                                    )
-@define_diffrule NaNMath.log2(x)   = :(  inv($x) / NaNMath.log(2)                   )
-@define_diffrule NaNMath.log10(x)  = :(  inv($x) / NaNMath.log(10)                  )
+@define_diffrule NaNMath.log2(x)   = :(  inv($x) / NaNMath.log(oftype($x, 2))       )
+@define_diffrule NaNMath.log10(x)  = :(  inv($x) / NaNMath.log(oftype($x, 10))      )
```
devmotion, line 226: Couldn't we just use
```suggestion
@define_diffrule NaNMath.log2(x)   = :(  inv($logtwo * $x)  )
@define_diffrule NaNMath.log10(x)  = :(  inv($logten * $x)  )
```
or something similar here?

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
+            @test DiffRules.hasdiffrule(M, f, 1)
+            deriv = DiffRules.diffrule(M, f, :goo)
+            modifier = if f in (:asec, :acsc, :asecd, :acscd, :acosh, :acoth)
+                1.0
+            elseif f === :log1mexp
+                -1.0
+            elseif f === :log2mexp
+                -0.5
+            else
+                0.0
```
devmotion, line 31: The modifiers should be of type `T` as well, I assume.

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
-        derivs = DiffRules.diffrule(M, f, :foo, :bar)
-        @eval begin
-            let
-                if "mod" == string($M.$f)
-                    foo, bar = rand() + 13, rand() + 5 # make sure x/y is not integer
-                else
-                    foo, bar = rand(1:10), rand()
+            @eval begin
+                let
+                    goo = $T(rand() + $modifier)
```
devmotion, line 35: Ah you account for `T` here... Instead you could account for the type of the modifier above and use
```suggestion
                    goo = rand($T) + $modifier
```
here

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
+                    @test isapprox($deriv, finitediff($M.$f, goo), rtol=0.05)
```
devmotion, line 36: Would be good to check that `$deriv` is of type `T`, I assume?
```suggestion
                    @test $deriv isa $T
                    @test isapprox($deriv, finitediff($M.$f, goo), rtol=0.05)
```

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
+                    # test for 2pi functions
+                    if "mod2pi" == string($M.$f)
+                        goo = 4pi + $modifier
+                        @test NaN === $deriv
```
devmotion, line 40: This is always `Float64` - is this desired?

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
-                if !(isnan(dy))
-                    @test isapprox(dy, finitediff(z -> $M.$f(foo, z), bar), rtol=0.05)
+            end
+        elseif arity == 2
+            @test DiffRules.hasdiffrule(M, f, 2)
+            derivs = DiffRules.diffrule(M, f, :foo, :bar)
+            @eval begin
+                let
+                    if "mod" == string($M.$f)
+                        foo, bar = $T(rand() + 13), $T(rand() + 5) # make sure x/y is not integer
```
devmotion, line 50:
```suggestion
                        foo, bar = rand($T) + 13, rand($T) + 5 # make sure x/y is not integer
```

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
+                    else
+                        foo, bar = rand(1:10), $T(rand())
```
devmotion, line 52:
```suggestion
                        foo, bar = rand(1:10), rand($T)
```

test/runtests.jl
```diff
@@ -16,67 +16,69 @@ end
...
+                    end
+                    dx, dy = $(derivs[1]), $(derivs[2])
```
devmotion, line 54: Same here check the type of `dx` and `dy`?

src/rules.jl
torfjelde: That's a typo :upside_down_face:

test/runtests.jl
torfjelde: I swear I had this earlier but messed up the merge or something. Anyways, I'm def with you that this is a good idea :+1:

src/rules.jl
```diff
@@ -217,8 +222,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule NaNMath.acosh(x)  = :(  inv(NaNMath.sqrt(NaNMath.pow($x, 2) - 1))  )
 @define_diffrule NaNMath.atanh(x)  = :(  inv(1 - NaNMath.pow($x, 2))                )
 @define_diffrule NaNMath.log(x)    = :(  inv($x)                                    )
-@define_diffrule NaNMath.log2(x)   = :(  inv($x) / NaNMath.log(2)                   )
-@define_diffrule NaNMath.log10(x)  = :(  inv($x) / NaNMath.log(10)                  )
+@define_diffrule NaNMath.log2(x)   = :(  inv($(DiffRules.logtwo) * $x)  )
+@define_diffrule NaNMath.log10(x)  = :(  inv($(DiffRules.logten) * $x)  )
```
torfjelde, line 226: @devmotion I remembered why I did this: https://github.com/mlubin/NaNMath.jl/issues/47

src/DiffRules.jl
```diff
@@ -2,6 +2,8 @@ __precompile__()
 
 module DiffRules
 
+using IrrationalConstants
```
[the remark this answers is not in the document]
mcabbott: Correct!

What's the argument here against something like `:( inv($x) / oftype(float($x), log(10)) )`, and `:($x / 2 / pi)`?

test/runtests.jl
torfjelde: It's a bit more difficult to easily address because `foo` is sometimes an integer, e.g. what do we do for `*`?

src/DiffRules.jl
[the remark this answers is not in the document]
devmotion: I would again argue, using irrationals is cleaner, simpler, and avoids unnecessary loss of precision if desired. One can avoid heuristics such as `float($x)` and instead computes exactly the desired result in the most natural and least invasive types. Moreover, e.g., `log(10)` will always be of type `Float64` and hence the reduced precision can't be recovered when working with other more accurate number types.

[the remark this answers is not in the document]
devmotion: IrrationalConstants is also already an indirect dependency of DiffRules, both through SpecialFunctions and LogExpFunctions, so adding it as an dependency does not affect the tree of dependencies.

devmotion:
> Doing stuff like $logten instead of $(DiffRules.logten) won't work unless you also have done using IrrationalConstants in the scope where you use eval. I recall now that this was the original reason why I added the module-specification.

This won't work though - DiffRules is not necessarily available in the module where you `eval` the rules. For instance, it seems it would fail if you just load `diffrules` with `using DiffRules: diffrules` (and this example doesn't seem too far fetched).

torfjelde:
> > Doing stuff like $logten instead of $(DiffRules.logten) won't work unless you also have done using IrrationalConstants in the scope where you use eval. I recall now that this was the original reason why I added the module-specification.
> 
> This won't work though - DiffRules is not necessarily available in the module where you `eval` the rules. For instance, it seems it would fail if you just load `diffrules` with `using DiffRules: diffrules` (and this example doesn't seem too far fetched).

Sure, but then we can't use this approach using irrationals?

devmotion:
I still think it works just with standard interpolation, I'm not sure why it failed when you tried it previously (maybe you use e.g. `:($x + logten)` instead of `:($x + $logten)`?):
```julia
julia> module A
           f(x) = x
           g(x) = :($x + $f($x))
       end
Main.A

julia> module B
           using ..A: g
           @eval begin
               r(x) = $(g(:x))
           end
       end
Main.B

julia> B.r(1.0)
2.0
```

devmotion:
Or with IrrationalConstants:
```julia
julia> module A
           using IrrationalConstants: logtwo
           g(x) = :($x + $logtwo)
       end
Main.A

julia> module B
           using ..A: g
           @eval begin
               r(x) = $(g(:x))
           end
       end
Main.B

julia> B.r(1.0)
1.6931471805599454
```

src/DiffRules.jl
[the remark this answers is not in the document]
mcabbott: Yes I agree it looks cleaner. I worry, as before, about going from loading nothing to loading packages... which may decide they want to load a few other small packages... 

One thought is this: If SpecialFunctions loads this, then calling things as `SpecialFunctions.IrrationalConstants.twopi` is guaranteed to work, right?

devmotion:
>  it looks like there's no reference to the original module, etc. So how does it then manage to resolve it back to where the irrational is defined?

It's just not displayed but it is actually a proper reference to the irrational.

> Btw, though I agree that this is nicer, you still need to have DiffRules available in the scope when doing eval for general rules, e.g.

It seems if `_abs_deriv` is interpolated it's not necessary anymre that DiffRules is available.

torfjelde:
> It's just not displayed but it is actually a proper reference to the irrational.

Well, that's confusing :sweat_smile: But why isn't this the case when I do `DiffRules.logten` then? Shouldn't this also be the same?

> It seems if _abs_deriv is interpolated it's not necessary anymre that DiffRules is available.

Oh of course!

src/DiffRules.jl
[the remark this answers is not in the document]
devmotion: Since it's also a dependency of LogExpFunctions, the choice is a bit arbitrary... And in any case it would require that `SpecialFunctions` is available in the module where the rule is evaled - which might not be the case, e.g., if you are only interested in defining rules for functions from Base (it's easy to filter them, even more so with the new keyword argument for DiffRules).

I think this example highlights that in fact it might be better to interpolate all such references in the expressions - to irrationals but also to functions from e.g. SpecialFunctions (and hence to load SpecialFunctions). It's less likely that evaluating rules leads to undefined references and hence seems much safer in general. I'm still not convinced that avoiding `using IrrationalConstants` or `using SpecialFunctions` is really desirable - if these packages are broken most downstream packages that use DiffRules will be broken anyway (and these packages are so central that it would affect large parts of the Julia ecosystem) and hence I don't think one has to be concerned with possible upstream issues when loading the packages. Whereas I think DiffRules should be concerned with issues caused by non-interpolated local references in the rules.

devmotion:
> Shouldn't this also be the same?

Yes, but it's unnecessarily verbose. I was confused initially because it sounded like you used it since `$logten` did not work - but since both approaches result in the same expressions this did not make sense to me.

torfjelde:
> > Shouldn't this also be the same?
> 
> Yes, but it's unnecessarily verbose. I was confused initially because it sounded like you used it since `$logten` did not work - but since both approaches result in the same expressions this did not make sense to me.

Yeah, it was because I only looked at the expanded expression and noted it had lost the reference. Sorry, not making that mistake again:) Thanks man!

src/DiffRules.jl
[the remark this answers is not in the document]
torfjelde: Any updates on this?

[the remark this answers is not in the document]
mcabbott: My opinion is that we should change old packages like this as little as possible, at least not without serious gain. E.g. "interpolate all such references" might have been a better design, but isn't the one chosen, and changing all of them sounds like a way to create unexpected headaches, and gain nothing.

I also don't think it should go from loading zero dependencies to loading a bunch, without serious gain. IIRC there are exactly 4 cases where writing `SpecialFunctions.IrrationalConstants.sqrtπ` would not work, and they matter only if you happen to want to AD a few things like `log10(::BigFloat)` efficiently. That seems pretty obscure.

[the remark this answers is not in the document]
torfjelde: > and changing all of them sounds like a way to create unexpected headaches, and gain nothing.

I would argue that the gain is not nothing. I have code that runs 2X faster by using this branch vs. previous release when using Zygote (it's a pretty high-dim problem). And possibly even more significant: last time I checked the lack of proper conversion completely breaks broadcasting over `ForwardDiff.Dual` on GPU, which also means that Zygote breaks when you try to broadcast over some of these functions.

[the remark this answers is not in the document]
mcabbott: No, sorry, I don't mean that types don't matter. I mean that changing from an Expr containing a Symbol, to one containing a Function, doesn't matter, as the `@eval`'d code ends up the same.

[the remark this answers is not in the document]
torfjelde: Ah, gotcha :+1:

> I don't see any such changes in this PR right now.

So you're happy with the `$logten`, etc. that I've done but don't want to start interpolating _everything_?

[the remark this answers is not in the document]
torfjelde: And the only issue that is holding back this PR atm is whether we should depend directly or indirectly on `IrrationalConstants`?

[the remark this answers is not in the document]
mcabbott: I'm against adding dependencies. I believe that you can have all the benefits of `sqrtπ` without explicitly loading the package, which you must do to interpolate `:($sqrtπ)` from this package's scope. 

It seems reasonable to assume that any user of the SpecialFunctions rules must have SpecialFunctions loaded, and thus I don't see a problem with accessing constants within it: I find it hard to imagine where `:( 2 * (($x * SpecialFunctions.erfcx($x)) - $invsqrtπ) )` would work but something like  `:( 2 * (($x * SpecialFunctions.erfcx($x)) - SpecialFunctions.erfcx.invsqrtπ) )` would not. 

I don't think `:( inv($logten * $x) )` is worth a dependency, compared to `:( inv($logten * log(oftype($x, 10))) )`.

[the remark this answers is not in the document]
devmotion: I think interpolating everything is the cleaner and safer way but out of scope for this PR. IMO, however, we should load and interpolate the irrational constants instead of relying on workarounds that are trivial to break and are inconsistent with our API that allows to filter rules for specific packages if only a subset of them is loaded.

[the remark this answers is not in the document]
mcabbott: How is this inconsistent with the API? If you have loaded SpecialFunctions, then `@eval` will work. If not, not.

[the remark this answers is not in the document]
devmotion: It is inconsistent if we use SpecialFunctions.IrrationalConstants... in rules for functions that are not defined in SpecialFunctions.

I am very strongly against SpecialFunctions.IrrationalConstants... since it is a hack without IMO clear benefit but sever disadvantages: If SpecialFunctions decides to drop the dependency on IrrationalConstants, which can be done in any non-breaking release, then DiffRules and all downstream packages will be broken. And in fact, this PR will break all downstream packages and uses for SpecialFunctions < 1.7.0 since only in this release the dependency on IrrationalConstants was introduced in SpecialFunctions. Ie we would have to bump the compat without any other reason, and could still not be sure if SpecialFunctions 2.1.0 breaks the whole DiffRules ecosystem.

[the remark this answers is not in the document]
mcabbott: > in rules for functions that are not defined in SpecialFunctions.

Yes, there are 4 such, which only matter for some strange BigFloat case.

> If SpecialFunctions decides to drop the dependency on IrrationalConstants, which can be done in any non-breaking release

I agree this trick upgrades that to part of SpecialFunctions's API, unavoidably.

Dropping SpecialFunctions 0.10 seems like no problem. Why is a lower bound at 1.7.0 going to break anything?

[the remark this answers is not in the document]
torfjelde: > Yes, you can't do that. There are 4 such in the PR, which only matter for some strange BigFloat case, and can just be reverted.

Am a bit confused by this statement. If we remove the usage of irrational constants, won't the `log10` rule result in promotion to `Float64` from `Float32`? This doesn't seem like a "strange case".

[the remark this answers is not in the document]
mcabbott: The alternative to a fancy log10 is `log(oftype(x, 10)`. Which compiles to the exact same code. Except, I think, for `x::BigFloat` which is mutable. (But slow in any case, and probably has not occurred in the wild.)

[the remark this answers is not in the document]
torfjelde: Ah gotcha :+1: 

I'm still confused as to why it's preferable to depend on IrrationalConstants implicitly vs. explicitly, given that explicit dependence allows us to control our dependence on IrrationalConstants directly (and thus avoiding issues/annoyances as outlined by @devmotion above) rather than relying on SpecialFunctions not doing something surprising.

> I agree this trick upgrades that to part of SpecialFunctions's API, unavoidably.

But isn't it avoidable by just depending on IrrationalConstants directly?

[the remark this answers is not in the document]
mcabbott: The point of the trick is to avoid *loading* any packages. That's one meaning of dependency. And it's the meaning under which Static.jl managed to break about a thousand packages which didn't know they depended on it, last year. That's the sort of britleness I'd like to avoid having more of. I suspect this is a losing battle, and soon everything will depend on everything else, but I try.

This package still *constrains* SpecialFunctions via compat, which is read by Pkg without loading any code. That's another meaning of dependency, and would have avoided the catastrophe. This package could similarly constrain IrrationalConstants if this seems like a good idea. Adding that would be free, since (for suitably new SpecialFunctions) Pkg already downloads the package.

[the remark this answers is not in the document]
devmotion: I disagree, I think making it an implicit dependency makes DiffRules very brittle because every new SpecialFunctions release may break DiffRules without us having any control over it. That seems much more brittle than potential issues with `using IrrationalConstants`. IrrationalConstants is such a central package in the ecosystem that I don't think DiffRules can avoid downstream errors by not loading it if e.g. SpecialFunctions, LogExpFunctions, StatsFuns etc all depend on it and load it.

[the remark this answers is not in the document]
torfjelde: Oooh I didn't realize DiffRules no longer loads SpecialFunctions.. The discussion makes more sense now :+1: 

Well, though I'm still uncertain which approach is preferred, I'm quite keen on getting this through, and so I'm fine with what you're suggesting :+1: 

Btw, is there a particular reason why we shouldn't just move the rules for SpecialFunctions into, well, SpecialFunctions itself?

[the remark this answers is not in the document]
torfjelde: I think @mcabbott 's point is that diffrules for SpecialFunctions will only be evaluated in a scope where SpecialFunctions is available, and so if indeed a SpecialFunctions release breaks the diffrules for methods from SpecialFunctions, doing `using DiffRules` will still work if you're not calling `@eval` on any of these particular rules.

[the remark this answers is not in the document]
devmotion: > I think @mcabbott 's point is that diffrules for SpecialFunctions will only be evaluated in a scope where SpecialFunctions is available, and so if indeed a SpecialFunctions release breaks the diffrules for methods from SpecialFunctions, doing `using DiffRules` will still work if you're not calling `@eval` on any of these particular rules. 
> 
> I.e. SpecialFunctions will only break rules for SpecialFunctions.

Sure, but how is this not the responsibility of DiffRules which defines these brittle rules in the first place? In such a scenario the only proper fix is to change the definition of the rule in DiffRules. I still don't understand why we should risk breaking the whole AD ecosystem instead of loading a lightweight dependency that is loaded already by AD packages but also other core packages, directly or indirectly.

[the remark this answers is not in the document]
mcabbott: > Btw, is there a particular reason why we shouldn't just move the rules for SpecialFunctions into, well, SpecialFunctions itself?

I think this might be a limitation of the design of ForwardDiff (and perhaps others). The `eval` code on loading, and never look again, so rules which aren't in the list at first won't be loaded. Certainly that's true for independent packages loaded later (ChainRules is invented partly to solve this problem) but there's a chance that rules defined in SpecialFunctions would make it, so long as `using SpecialFuncitons` is above the `eval` loop in ForwardDiff? Haven't tried. Might break other things.

> SpecialFunctions will only break rules for SpecialFunctions

Yes. Downstream users of ForwardDiff who only want Base functions need never know if things are (temporarily) broken, because defining a function which calls a nonexistent symbol isn't an error (until you call it).

I guess the chance of IrrationalFunctions causing a Static.jl-style apocaplyse are low, since it's much smaller, simpler. Although I thought this about EllipsisNotation, too, which had been the same 5 lines (zero dependencies) for years and years, and then somehow acquired a dependence which had dependants and suddenly it's many thousands of lines and Base.Experimental macros and all sorts of weird highly optimised stuff. 

Anyway, I have probably said too much, and should shut up. Maybe I unsubscribe. If this PR does want to go from zero to one full-time dependencies, I do think that's something another maintainer ought to sign off on. Sorry if this is more of a rat's nest than you wanted to get into, Tor, thanks for the work.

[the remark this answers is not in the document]
torfjelde: > I think this might be a limitation of the design of ForwardDiff (and perhaps others). The eval code on loading, and never look again, so rules which aren't in the list at first won't be loaded.

Yeah I just remembered how ForwardDiff's mechanism works. If packages were to define their own rules, they'd have to also load ForwardDiff.

> Yes. Downstream users of ForwardDiff who only want Base functions need never know if things are (temporarily) broken, because defining a function which calls a nonexistent symbol isn't an error (until you call it).

But are there many packages that use DiffRules in this way? Maybe SymbolicUtils.jl? But most dependance on DiffRules comes through ForwardDiff, which already loads SpecialFunctions, etc.

> I guess the chance of IrrationalFunctions causing a Static.jl-style apocaplyse are low, since it's much smaller, simpler. Although I thought this about EllipsisNotation, too, which had been the same 5 lines (zero dependencies) for years and years, and then somehow acquired a dependence which had dependants and suddenly it's many thousands of lines and Base.Experimental macros and all sorts of weird highly optimised stuff.

These experiences with Static.jl and EllipsisNotation.jl you're talking about, was this prior to compat-entries in Package.toml? 

> If this PR does want to go from zero to one full-time dependencies, I do think that's something another maintainer ought to sign off on. 

As I've said, I'm happy to go with the zero dependency approach now that I've realized this is the path DiffRules has already gone down. Whether or not this should be the case, should maybe be left for as a separate issue/PR?

> Sorry if this is more of a rat's nest than you wanted to get into, Tor, thanks for the work.

No worries! I didn't have expectations of this being just a "quick patch" exactly due to how much code depends on DiffRules. And thank you for explaining everything in great detail and being so patient with me :+1: 

_So_. How about we take the approach of _not_ loading IrrationalConstants _in this PR_, and defer introduction of loading IrrationalConstants to a separate issue/PR? I think the fixes this PR has is of sufficient value that it's worth just getting it in "ASAP".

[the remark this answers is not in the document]
torfjelde: Okay, so one problem: SpecialFunctions.jl only imports the variables it explicitly needs https://github.com/JuliaMath/SpecialFunctions.jl/blob/414385a7097e0f376b1273a3e0deac9d30cf0f9e/src/SpecialFunctions.jl#L3-L14

This in turn means that we can't just do `SpecialFunctions.IrrationalConstants.sqrtπ`, but instead have to do `SpecialFunctions.sqrtπ`. Now this is fine _for now_, but will break if SpecalFunctions decides to change this and/or we need some constants that aren't explicitly used in SpecialFunctions.

src/rules.jl
```diff
@@ -211,14 +216,14 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule NaNMath.sqrt(x)   = :(  inv(2 * NaNMath.sqrt($x))                  )
 @define_diffrule NaNMath.sin(x)    = :(  NaNMath.cos($x)                            )
 @define_diffrule NaNMath.cos(x)    = :( -NaNMath.sin($x)                            )
-@define_diffrule NaNMath.tan(x)    = :(  1 + NaNMath.pow(NaNMath.tan($x), 2)        )
-@define_diffrule NaNMath.asin(x)   = :(  inv(NaNMath.sqrt(1 - NaNMath.pow($x, 2)))  )
-@define_diffrule NaNMath.acos(x)   = :(  -inv(NaNMath.sqrt(1 - NaNMath.pow($x, 2))) )
-@define_diffrule NaNMath.acosh(x)  = :(  inv(NaNMath.sqrt(NaNMath.pow($x, 2) - 1))  )
-@define_diffrule NaNMath.atanh(x)  = :(  inv(1 - NaNMath.pow($x, 2))                )
+@define_diffrule NaNMath.tan(x)    = :(  1 + NaNMath.pow(NaNMath.tan($x), oftype($x, 2))        )
```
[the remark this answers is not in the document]
torfjelde: You mean

```suggestion
@define_diffrule NaNMath.tan(x)    = :(  1 + NaNMath.tan($x)^2        )
```

If so, sure! I assumed `NaNMath.pow` was there for reason, and I needed to do `oftype` because otherwise the call ended up in stack overflow.

src/DiffRules.jl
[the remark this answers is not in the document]
devmotion: > As I've said, I'm happy to go with the zero dependency approach now that I've realized this is the path DiffRules has already gone down. Whether or not this should be the case, should maybe be left for as a separate issue/PR?

DiffRules also loaded e.g. LogExpFunctions in some releases (to figure out if some functions are available), this was changed only recently. I think there's a significant difference between both packages even though both are used by SpecialFunctions - IrrationalConstants is much more central and lightweight and only adds constants that are not in base (by committing type piracy technically but this seems OK - and definitely better than defining these constants in multiple packages since then you end up with many method redefinition warnings). Hence as an extension of base it's less surprising that it is useful even for derivatives of functions from base.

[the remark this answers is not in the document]
torfjelde: >  I think there's a significant difference between both packages even though both are used by SpecialFunctions - IrrationalConstants is much more central and lightweight and only adds constants that are not in base (by committing type piracy technically but this seems OK - and definitely better than defining these constants in multiple packages since then you end up with many method redefinition warnings). Hence as an extension of base it's less surprising that it is useful even for derivatives of functions from base.

Just for the record: I'm also in favour of loading IrrationalConstants.jl, but my opinion isn't so important here given that I'm not one of the maintainers of DiffRules :shrug: At this point, I just want to get this PR merged. _But_, as I pointed out above, the "naive" way of just using whatever is present in SpecialFunctions seems _very_ brittle.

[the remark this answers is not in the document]
devmotion: I guess @mcabbott and I both stated and explained our preferences, so maybe it would be good to include some other contributor. @andreasnoack, it seems you commented on an earlier draft of this PR - what's your opinion? Should we depend on, load IrrationalConstants, and interpolate constants or refer to (unexported) constants in SpecialFunctions without depending and loading IrrationalConstants (and use them only for functions from SpecialFunctions)?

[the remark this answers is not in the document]
torfjelde: Just checking up on this again. Any conclusion to this? Are we still waiting for another maintainer to respond or has it already been decided that it's a "nah"?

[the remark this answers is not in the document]
devmotion: Since nobody has replied, let's just do it in the controlled future-proof way and load and interpolate the constants. Since we already depend on it indirectly, it shouldn't cause any problems (BTW in the mean time also CR added it as a dependency). And it's easy to change if we want to in the future and there's no risk of SpecialFunctions breaking anything accidentally.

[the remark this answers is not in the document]
torfjelde: That is, as it's currently implemented? Sooo we merge?

Project.toml
```diff
@@ -3,20 +3,23 @@ uuid = "b552c78f-8df3-52c6-915a-8e097449b14b"
 version = "1.9.0"
```
torfjelde, line 3:
```suggestion
version = "1.9.1"
```

src/rules.jl
```diff
@@ -5,46 +5,46 @@
 # unary #
 #-------#
 
-@define_diffrule Base.:+(x)                   = :(  1                                  )
-@define_diffrule Base.:-(x)                   = :( -1                                  )
+@define_diffrule Base.:+(x)                   = :(  one($x)                            )
+@define_diffrule Base.:-(x)                   = :( -one($x)                            )
```
devmotion, line 9: Is this needed? The derivative does not necessarily have exactly the same typr as `x`, I guess we mainly want to avoid unnecessary promotions. But `1` and `-1` already seem quite safe and shouldn't cause any promotions?

src/rules.jl
```diff
@@ -5,46 +5,46 @@
...
-@define_diffrule Base.tand(x)                 = :(  (π / 180) * (1 + tand($x)^2)       )
-@define_diffrule Base.secd(x)                 = :(  (π / 180) * secd($x) * tand($x)    )
-@define_diffrule Base.cscd(x)                 = :( -(π / 180) * cscd($x) * cotd($x)    )
-@define_diffrule Base.cotd(x)                 = :( -(π / 180) * (1 + cotd($x)^2)       )
+@define_diffrule Base.sind(x)                 = :(  π * cosd($x) / 180                 )
+@define_diffrule Base.cosd(x)                 = :( -(π * sind($x) / 180)               )
+@define_diffrule Base.tand(x)                 = :(  π * (1 + tand($x)^2) / 180         )
+@define_diffrule Base.secd(x)                 = :(  π * secd($x) * tand($x) / 180      )
+@define_diffrule Base.cscd(x)                 = :( -(π * cscd($x) * cotd($x) / 180)    )
+@define_diffrule Base.cotd(x)                 = :( -(π * (1 + cotd($x)^2) / 180)       )
```
devmotion, line 33: Can we use `deg2rad` as in https://github.com/JuliaDiff/ChainRules.jl/pull/554?

```suggestion
@define_diffrule Base.sind(x)                 = :(  deg2rad(cosd($x))  )
@define_diffrule Base.cosd(x)                 = :( - deg2rad(sind($x))   )
@define_diffrule Base.tand(x)                 = :(  deg2rad(1 + tand($x)^2)  )
@define_diffrule Base.secd(x)                 = :(  deg2rad(secd($x) * tand($x))  )
@define_diffrule Base.cscd(x)                 = :( - deg2rad(cscd($x) * cotd($x))  )
@define_diffrule Base.cotd(x)                 = :( - deg2rad(1 + cotd($x)^2)  )
```

src/rules.jl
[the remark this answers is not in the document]
devmotion: I agree, I think @mcabbott's suggestion would be preferable. Why can't we lower bound NaNMath?

src/rules.jl
devmotion: Maybe we could even use `^ 2` instead of `pow` here?

test/runtests.jl
devmotion: We could only perform type checks if it's not an integer?

torfjelde: Yeah, already working on this locally:)

src/rules.jl
torfjelde: Sure, good point :+1: I'll revert this

src/rules.jl
```diff
@@ -88,8 +87,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
 @define_diffrule Base.ldexp(x, y)  = :( exp2($y)                                                ), :NaN
 
-@define_diffrule Base.mod(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -floor(z)), NaN)) )
-@define_diffrule Base.rem(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -trunc(z)), NaN)) )
+@define_diffrule Base.mod(x, y)    = :( z = $x / $y; ifelse(isinteger(z), oftype(z, NaN), one(z)) ), :(  z = $x / $y; ifelse(isinteger(z), oftype(z, NaN), -floor(z)) )
+@define_diffrule Base.rem(x, y)    = :( z = $x / $y; ifelse(isinteger(z), oftype(z, NaN), one(z)) ), :(  z = $x / $y; ifelse(isinteger(z), oftype(z, NaN), -trunc(z)) )
```
torfjelde, line 91: I'm honestly a bit confused as to why we had the `first(promote` there before. Got any clue @devmotion ?

test/runtests.jl
torfjelde: Added.

src/rules.jl
devmotion: I don't know :shrug: I'm just slightly worried about the variable definitions, I guess `z` might be defined twice in downstream implementations and could possibly leak (IIRC there were some problems even with the tests here at some point and hence I introduced the `let` blocks in the tests). `z` is defined in the current implementation as well, so it seems to be fine. Alternatively, one could use something like `let z = $x / $y; ....; end` but I'm not sure if it's needed.

src/rules.jl
```diff
@@ -112,18 +111,24 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
...
 # unary #
 #-------#
-
-@define_diffrule SpecialFunctions.erf(x)         = :(  (2 / sqrt(π)) * exp(-$x * $x)       )
+@define_diffrule SpecialFunctions.erf(x) = :( 2 * exp(-$x^2) * $invsqrtπ )
+@define_diffrule SpecialFunctions.erfinv(x) = :( (exp(SpecialFunctions.erfinv($x)^2) * $sqrtπ) / 2 )
+@define_diffrule SpecialFunctions.erfc(x) = :( -(exp(-$x * $x) * $invsqrtπ) * 2 )
+@define_diffrule SpecialFunctions.erfcinv(x) = :( -(exp(SpecialFunctions.erfcinv($x)^2) * $sqrtπ) / 2 )
+@define_diffrule SpecialFunctions.erfi(x) = :( 2 * exp($x * $x) * $invsqrtπ )
+@define_diffrule SpecialFunctions.erfcx(x) = :( 2 * ($x * SpecialFunctions.erfcx($x) - $invsqrtπ) )
```
devmotion, line 119: Are there any new rules? Or did you just move them around?

src/rules.jl
```diff
@@ -162,19 +167,19 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 #--------#
 
 @define_diffrule SpecialFunctions.besselj(ν, x)   =
-    :NaN, :(  (SpecialFunctions.besselj($ν - 1, $x) - SpecialFunctions.besselj($ν + 1, $x)) / 2  )
+    :(oftype(float($ν), NaN)), :(  (SpecialFunctions.besselj($ν - 1, $x) - SpecialFunctions.besselj($ν + 1, $x)) / 2  )
```
devmotion, line 170: Is this needed? If this partial is included in any computations it will break them anyway.

src/rules.jl
torfjelde: Oh crap, I think these are repeated by a mistake when merging with the main branch or something.

src/rules.jl
torfjelde: Up to you tbh. I like the consistency, but I'm also fine with not doing it (just need to filter these during testing).

devmotion: Hm I think I prefer if we keep `NaN` if there are no clear reasons to change it. `NaN` seems a bit simpler and in DiffRules usually indicates if a derivative does not exist or is not implemented. There are also some other rules that still return `:NaN` in this PR. My suggestion would be to keep `:NaN` and ensure type stability and avoid promotions in cases where only sometimes `NaN` is returned (e.g., `isinteger(x) ? NaN : 1`).

src/rules.jl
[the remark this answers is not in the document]
torfjelde: Exactly

src/rules.jl
torfjelde: True, I just went with it despite that because the current implementation had `z` in there.. But what do you prefer me doing here? Should we leave as is or introduce `let`?

src/rules.jl
torfjelde: Aight, will do :+1: I'll just filter them out from the tests then.

src/rules.jl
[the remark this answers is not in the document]
devmotion: Great, then it's easiest to revert the changes here and possible switch to `^2` in a separate PR :slightly_smiling_face:

src/rules.jl
devmotion: It seems @mcabbott also mentioned it above and considered a gensym alternative (`let` might be easier to debug and inspect). I agree with your comment there and think probably this should be addressed in a separate PR (if we want to change anything). I'm fine with the `z` in this PR.

I noticed another potential issue though: `z` is not necessarily a floating point number, e.g., if `x` and `y` are `Rational`s. Maybe use `float(z)` instead?

test/runtests.jl
```diff
@@ -1,82 +1,117 @@
...
-end
+# Set `max_range` to avoid domain errors.
+const finitediff = central_fdm(5, 1, max_range=1e-3)
 
 @testset "DiffRules" begin
 @testset "check rules" begin
 
 non_diffeable_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ldexp, 2), (:Base, :ifelse, 3)]
 
 for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
```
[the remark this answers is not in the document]
devmotion: Seems useful.

src/rules.jl
torfjelde: Ah yes, good catch :+1: I noticed this bug in the bessel functions, etc. but forgot to add it here.

src/rules.jl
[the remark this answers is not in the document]
torfjelde: Oh I guess I spoke a bit too sone in my previous reply: if we revert, we need to lower-bound NaNMath to avoid the infinite-recursion issue :confused: 

So we have two alternatives:
1. Revert and lower-bound NaNMath.
2. Don't revert and don't lower-bound NaNMath.

test/runtests.jl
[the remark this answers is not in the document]
torfjelde: Want me to add it?

Project.toml
```diff
@@ -1,22 +1,25 @@
 name = "DiffRules"
 uuid = "b552c78f-8df3-52c6-915a-8e097449b14b"
-version = "1.9.0"
+version = "1.9.1"
```
torfjelde, line 3: Is this a breaking change @devmotion ?

src/rules.jl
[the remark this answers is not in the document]
devmotion: But the bug already exists with the master branch? And is not caused by DiffRules and fixed upstream? If we would revert and lower bound NaNMath, then users with an older version will just get an older version of DiffRules with the same bug, don't they?

test/runtests.jl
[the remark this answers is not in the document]
devmotion: Yes :+1:

Project.toml
devmotion: It shouldn't, it should just fix `Float32` etc. issues. Did I miss anything breaking?

torfjelde: No I agree, but the discussion regarding including a dependency made me worried and so I was thinking maybe we should make it breaking just to be on the safe-side. But _I_ would personally decide to make it non-breaking; just wanted to check with one of you.

src/rules.jl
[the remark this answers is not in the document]
torfjelde: > If we would revert and lower bound NaNMath, then users with an older version will just get an older version of DiffRules with the same bug, don't they?

Ah, very valid point yeah :+1: I'll revert!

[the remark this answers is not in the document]
torfjelde: Ehhm unfortunately we can't revert this and still preserve the types. The fix I added to NaNMath was just to fix the infinite recursion issue, not a promotion issue. And because it's calling an external C library, we need the two arguments to be of the same floating point precision (AFAIK), hence we need to call `float` on each argument separately, leading to something like `NaNMath.pow($x, 2)` eventually calling `NaNMath.pow(promote(float($x), float(2))...)` which results in both being promoted to `Float64` even if `x` is `Float32`.

And regarding using literal `pow`, I have no clue if this is the right thing to do or not, but I assumed it's in NaNMath because there's a possibility of it _not_ handling `NaN` correctly?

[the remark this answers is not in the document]
devmotion: > And because it's calling an external C library, we need the two arguments to be of the same floating point precision (AFAIK), hence we need to call float on each argument separately, leading to something like NaNMath.pow($x, 2) eventually calling NaNMath.pow(promote(float($x), float(2))...) which results in both being promoted to Float64 even if x is Float32.

This seems like an upstream issue? You shouldn't run into these issues if you use something like
```julia
NaNMath.pow(x::Real, y::Real) = NaNMath.pow(promote(x, y)...)
NaNMath.pow(x::T, y::T) where {T<:Real} = ccall(..., float(x), float(y)) # c call
```
I.e., one just has to make sure that the arguments are promoted before they are converted to floating point numbers. There are e.g. multiple places in SpecialFunctions where one has to deal with such issues and therefore uses e.g. `map(float, promote(x, y))`.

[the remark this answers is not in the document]
devmotion: https://github.com/mlubin/NaNMath.jl/pull/52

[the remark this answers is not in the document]
torfjelde: Yeah I was just thinking there should be a way of doing this consistently; awesome! Thanks man:)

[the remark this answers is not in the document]
devmotion: You should be able to revert to the current implementation, the fix in NaNMath was merged and released.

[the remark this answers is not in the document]
torfjelde: Done :+1:

test/runtests.jl
```diff
@@ -1,82 +1,120 @@
...
+                    elseif $(f === :log1mexp)
+                        rand($T) - one($T)
+                    elseif $(f in (:log2mexp, :erfinv))
+                        rand($T) - $T(0.5)
+                    else
+                        rand($T)
+                    end
+                    # We're happy with types with the correct promotion behavior, e.g.
+                    # it's fine to return `1` as a derivative despite input being `Float64`.
+                    @test one($T) * $deriv isa $T
```
devmotion, line 42: A bit more direct maybe:
```suggestion
                    @test promote_type(typeof($deriv), $T) === $T
```

src/rules.jl
```diff
@@ -88,8 +87,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
 @define_diffrule Base.ldexp(x, y)  = :( exp2($y)                                                ), :NaN
 
-@define_diffrule Base.mod(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -floor(z)), NaN)) )
-@define_diffrule Base.rem(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -trunc(z)), NaN)) )
+@define_diffrule Base.mod(x, y)    = :( z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), one(float(z))) ), :(  z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), -floor(float(z))) )
+@define_diffrule Base.rem(x, y)    = :( z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), one(float(z))) ), :(  z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), -trunc(float(z))) )
```
devmotion, line 91: Maybe a bit simpler:
```suggestion
@define_diffrule Base.mod(x, y)    = :( z = float($x / $y); isinteger(z) ? oftype(z, NaN) : one(z) ), :(  z = float($x / $y); isinteger(z) ? oftype(z, NaN) : -floor(z) )
@define_diffrule Base.rem(x, y)    = :( z = float($x / $y); isinteger(z) ? oftype(z, NaN) : one(z) ), :(  z = float($x / $y); isinteger(z) ? oftype(z, NaN) : -trunc(z) )
```

test/runtests.jl
```diff
@@ -1,82 +1,120 @@
...
+                        rand($T) + $T(0.5), rand($T)
+                    else
+                        rand($T), rand($T)
+                    end
+                    dx, dy = $(derivs[1]), $(derivs[2])
+                    if !(isnan(dx))
+                        @test dx ≈ finitediff(z -> $M.$f(z, bar), foo) rtol=1e-2 atol=1e-3
+
+                        # Check type, if applicable.
+                        if foo isa AbstractFloat
```
devmotion, line 77: Why is this check needed? Isn't the derivative `NaN` if this is not the case, which is already checked above?

test/runtests.jl
```diff
@@ -1,82 +1,120 @@
...
+                            if dx isa Complex
+                                @test dx isa Complex{$T}
+                            else
+                                @test dx isa $T
+                            end
```
devmotion, line 82: We don't have to be worried about promotions here?
```suggestion
                            @test promote_type(typeof(real(dx)), $T) === $T
```

test/runtests.jl
```diff
@@ -1,82 +1,120 @@
...
+                    if !(isnan(dy))
+                        @test dy ≈ finitediff(z -> $M.$f(foo, z), bar) rtol=1e-2 atol=1e-3
+
+                        # Check type, if applicable.
+                        if bar isa AbstractFloat
+                            if dy isa Complex
+                                @test dy isa Complex{$T}
+                            else
+                                @test dy isa $T
+                            end
```
devmotion, line 94:
```suggestion
                            @test promote_type(typeof(real(dy)), $T) === $T
```

test/runtests.jl
```diff
@@ -1,82 +1,120 @@
...
-                @test isapprox(dy, finitediff(y -> $M.$f(foo, y, goo), bar), rtol=0.05)
+            @test isapprox(dy, finitediff(y -> $M.$f(foo, y, goo), bar), rtol=0.05)
             end
             if !(isnan(dz))
-                @test isapprox(dz, finitediff(z -> $M.$f(foo, bar, z), goo), rtol=0.05)
+            @test isapprox(dz, finitediff(z -> $M.$f(foo, bar, z), goo), rtol=0.05)
+            end
             end
+            =#
         end
```
devmotion, line 117: Revert these changes?

test/runtests.jl
```diff
@@ -131,12 +169,12 @@ end
...
-        @test isapprox(dy, finitediff(y -> ifelse(foo, y, goo), bar), rtol=0.05)
-        @test isapprox(dz, finitediff(z -> ifelse(foo, bar, z), goo), rtol=0.05)
-    end
+@eval begin
+foo = $cond
+bar, gee = randn(2)
+dx, dy, dz = $(derivs[1]), $(derivs[2]), $(derivs[3])
+@test isapprox(dy, finitediff(y -> ifelse(foo, y, goo), bar), rtol=0.05)
+@test isapprox(dz, finitediff(z -> ifelse(foo, bar, z), goo), rtol=0.05)
+end
```
devmotion, line 178: Revert?

devmotion (commented):
Looks good, only some minor questions/suggestions left :slightly_smiling_face:

test/runtests.jl
torfjelde: A good point :+1:

src/rules.jl
devmotion: Hmm, some other rules use the ternary operator so I assumed it's fine. But maybe safer to not change too many things here.

test/runtests.jl
```diff
@@ -1,82 +1,112 @@
...
                 end
                 if !(isnan(dy))
-                    @test isapprox(dy, finitediff(z -> $M.$f(foo, z), bar), rtol=0.05)
+                    @test isapprox(dy, finitediff(y -> $M.$f(foo, y, goo), bar), rtol=0.05)
+                end
+                if !(isnan(dz))
+                    @test isapprox(dz, finitediff(z -> $M.$f(foo, bar, z), goo), rtol=0.05)
                 end
             end
+            =#
```
torfjelde, line 108: Reverted this change as you requested @devmotion but because it's now within the for-loop, it messed up the diff.

devmotion: A bit annoying but OK :man_shrugging:

test/runtests.jl
devmotion: Did you check if it's needed?

torfjelde: Oh crap, I forgot abou this. Pushing now.

torfjelde:
Btw, seems like the integration tests for Symbolics.jl are broken.

devmotion:
> Btw, seems like the integration tests for Symbolics.jl are broken.

Yeah, seems it's a Symbolics issue, the same errors show up on their master branch: https://github.com/JuliaSymbolics/Symbolics.jl/runs/4975659103?check_suite_focus=true

devmotion (approved):
LGTM, thanks @torfjelde!

PR #62, opened 2021-07-05
Revert "comment out trinary rules"

mcabbott:
This reverts #60, thus un-reverts #54. 

Wants to be run with #61 and https://github.com/JuliaDiff/ForwardDiff.jl/pull/530 before merging.


devmotion (commented):
I didn't follow the original ForwardDiff issues and registry problems closely - is this guaranteed to not break any existing releases of ForwardDiff?

mcabbott:
Thanks for looking. I thought so but indeed we should re-check. 

ForwardDiff v0.10.18 and below have the error if there is a trinary rule. This is the line upper-bounding their DiffRules in the registry:

https://github.com/JuliaRegistries/General/blob/master/F/ForwardDiff/Compat.toml#L50-L51

The current version is ForwardDiff v0.10.19 which accepts only DiffRules 1.2.1 and up:

https://github.com/JuliaDiff/ForwardDiff.jl/blob/v0.10.19/Project.toml

This is reflected here in the registry:

https://github.com/JuliaRegistries/General/blob/master/F/ForwardDiff/Compat.toml#L22-L23

lines added by the robot https://github.com/JuliaRegistries/General/pull/41858 . So future releases should also work smoothly I believe, not touch old bounds.

So I think that implies this is safe.

src/rules.jl
```diff
@@ -100,15 +100,11 @@ end
 # trinary #
 #---------#
 
-#=
-
 @define_diffrule Base.muladd(x, y, z) = :($y), :($x), :(one($z))
 @define_diffrule Base.fma(x, y, z)    = :($y), :($x), :(one($z))
 
 @define_diffrule Base.ifelse(p, x, y) = false, :($p), :(!$p)
```
devmotion, line 106: Ideally, the first argument should be marked as non-differentiable I think. But AFAICT this is not supported by DiffRules (or rather handled in a crude way by returning `NaN` for both non-differentiable arguments and non-implemented derivatives)?

test/runtests.jl
```diff
@@ -82,7 +80,6 @@ for xtype in [:Float64, :BigFloat, :Int64]
 end
 
 # Test ifelse separately as first argument is boolean
-#=
 @test DiffRules.hasdiffrule(:Base, :ifelse, 3)
 derivs = DiffRules.diffrule(:Base, :ifelse, :foo, :bar, :goo)
 for cond in [true, false]
```
devmotion, line 85:
```suggestion
for cond in (true, false)
```

src/rules.jl
mcabbott: Yes I think I wasn't sure a year ago what this package did, but I see some other NaN. Which as you say seems a reasonable way to encode non-implemented / non-diff. 

```suggestion
@define_diffrule Base.ifelse(p, x, y) = NaN, :($p), :(!$p)
```

test/runtests.jl
```diff
@@ -94,4 +91,4 @@ for cond in [true, false]
         @test isapprox(dz, finitediff(z -> ifelse(foo, bar, z), goo), rtol=0.05)
```
devmotion, line 91: Maybe also check that `isnan(dx)`?

mcabbott:
```suggestion
        @test isapprox(dz, finitediff(z -> ifelse(foo, bar, z), goo), rtol=0.05)
        @test isnan(dx)
```

devmotion (approved):
Looks good to me.

devmotion:
@mcabbott I updated the PR and fixed some tests, do you have any comments? Otherwise I think this PR can be merged.

PR #66, opened 2021-08-27
Added some rules for scaled airy and bessel functions and incomplete gamma

amrods:


mcabbott:
Test failure on Julia 1.0:
```
Error During Test at /home/runner/work/DiffRules.jl/DiffRules.jl/test/runtests.jl:46
11
  Test threw exception
12
  Expression: isapprox(dy, finitediff((z->begin
13
                SpecialFunctions.gamma(foo, z)
14
            end), bar), rtol=0.05)
15
  MethodError: no method matching gamma(::Int64, ::Float64)
16
  Closest candidates are:
17
    gamma(::Union{Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8}) at /home/runner/.julia/packages/SpecialFunctions/ne2iw/src/gamma.jl:857
18
    gamma(::Real) at /home/runner/.julia/packages/SpecialFunctions/ne2iw/src/gamma.jl:561
19
  Stacktrace:
20
   [1] (::getfield(Main, Symbol("##60#62")))(::Float64) at ./none:0
21
   [2] finitediff(::getfield(Main, Symbol("##60#62")), ::Float64) at /home/runner/work/DiffRules.jl/DiffRules.jl/test/runtests.jl:15
22
```
Were these methods added to SpecialFunctions in a version which doesn't support Julia 1.0?

amrods:
Should I qualify some rules with something like
```julia
if VERSION < v"0.7-"
```
as you have for `atan2`?

mcabbott:
`atan2` was in Base a long long time ago, and doesn't exist anymore, really that should be deleted entirely.

For `gamma`, when were these methods added to SpecialFunctions?

amrods:
If you guide me to know when the method was added, I can do that. Same for the tests. :)

mcabbott:
Maybe it's enough to test `hasmethod(gamma, Tuple{Number, Number})` in the tests. 

It doesn't quite belong in `non_numeric_arg_functions` , and unlike `rem2pi` you don't really want a different test, you just want to skip the normal test if this doesn't exist. Perhaps start another list for this, and any other such, something like:
```
recently_defined_functions = []
if !hasmethod(SpecialFunctions.gamma, Tuple{Number, Number})
    # 2-arg gamma added in ??, 
    push!(recently_defined_functions, (...))
end
```
Re versions my thinking is that if in future this stops supporting some old versions of SpecialFunctions, then it's nice to know what bits you can delete. I looked through the closed PRs but this didn't jump out. I think the way to find out is git blame, which when viewing the source on github's web page is a button top-right.

amrods:
I installed Julia 1.0 to test locally and `SpecialFunctions` was downgraded to `v0.8.0`. In that version `gamma` only has methods for one argument.

amrods:
Now it's `ForwardDiff` that doesn't pass the tests :(

mcabbott:
Same test in Symbolics.jl failing on CI for an unrelated PR: https://github.com/JuliaSymbolics/Symbolics.jl/pull/373/checks?check_run_id=3563295246

Ditto for ModelingToolkit.jl failure: https://github.com/SciML/ModelingToolkit.jl/pull/1252/checks?check_run_id=3528813219

amrods:
So what can I do now?

src/rules.jl
```diff
@@ -141,6 +141,10 @@ end
     :(  SpecialFunctions.airybiprime($x)  )
 @define_diffrule SpecialFunctions.airybiprime(x) =
     :(  $x * SpecialFunctions.airybi($x)  )
+@define_diffrule SpecialFunctions.airyaix(x)     =
+    :(  sqrt($x) * SpecialFunctions.airyaix($x) + SpecialFunctions.airyaiprimex($x)  )
```
devmotion, line 145: I suggest using
```suggestion
    :(  SpecialFunctions.airyaiprimex($x) + sqrt($x) * SpecialFunctions.airyaix($x)  )
```
for consistency with the rule in SpecialFunctions: https://github.com/JuliaMath/SpecialFunctions.jl/blob/1febdd05902f5b3a3adc9915aba40b776aa3ea24/src/chainrules.jl#L19

src/rules.jl
```diff
@@ -170,29 +172,34 @@ end
     :NaN, :(  (SpecialFunctions.bessely($ν - 1, $x) - SpecialFunctions.bessely($ν + 1, $x)) / 2  )
 @define_diffrule SpecialFunctions.besselk(ν, x)   =
     :NaN, :( -(SpecialFunctions.besselk($ν - 1, $x) + SpecialFunctions.besselk($ν + 1, $x)) / 2  )
+@define_diffrule SpecialFunctions.besselkx(ν, x)  =
+    :NaN, :( -(SpecialFunctions.besselkx($ν - 1, $x) - 2SpecialFunctions.besselkx($ν, $x) + SpecialFunctions.besselkx($ν + 1, $x)) / 2 )
```
devmotion, line 176: Slightly simpler and consistent with SpecialFunctions (https://github.com/JuliaMath/SpecialFunctions.jl/blob/1febdd05902f5b3a3adc9915aba40b776aa3ea24/src/chainrules.jl#L109):
```suggestion
    :NaN, :( -(SpecialFunctions.besselkx($ν - 1, $x) + SpecialFunctions.besselkx($ν + 1, $x)) / 2 + SpecialFunctions.besselkx($ν, $x)  )
```

src/rules.jl
```diff
@@ -170,29 +172,34 @@ end
...
 @define_diffrule SpecialFunctions.hankelh1(ν, x)  =
     :NaN, :(  (SpecialFunctions.hankelh1($ν - 1, $x) - SpecialFunctions.hankelh1($ν + 1, $x)) / 2  )
+@define_diffrule SpecialFunctions.hankelh1x(ν, x) =
+    :NaN, :(  (SpecialFunctions.hankelh1x($ν - 1, $x) - 2im*SpecialFunctions.hankelh1x($ν, $x) - SpecialFunctions.hankelh1x($ν + 1, $x)) / 2  )
```
devmotion, line 180: Same here (https://github.com/JuliaMath/SpecialFunctions.jl/blob/1febdd05902f5b3a3adc9915aba40b776aa3ea24/src/chainrules.jl#L123):
```suggestion
    :NaN, :(  (SpecialFunctions.hankelh1x($ν - 1, $x) - SpecialFunctions.hankelh1x($ν + 1, $x)) / 2 - im * SpecialFunctions.hankelh1x($ν, $x)  )
```

src/rules.jl
```diff
@@ -170,29 +172,34 @@ end
...
 @define_diffrule SpecialFunctions.hankelh2(ν, x)  =
     :NaN, :(  (SpecialFunctions.hankelh2($ν - 1, $x) - SpecialFunctions.hankelh2($ν + 1, $x)) / 2  )
+@define_diffrule SpecialFunctions.hankelh2x(ν, x)  =
+    :NaN, :(  (SpecialFunctions.hankelh2x($ν - 1, $x) + 2im*SpecialFunctions.hankelh2x($ν, $x) - SpecialFunctions.hankelh2x($ν + 1, $x)) / 2  )
```
devmotion, line 184: And here (https://github.com/JuliaMath/SpecialFunctions.jl/blob/1febdd05902f5b3a3adc9915aba40b776aa3ea24/src/chainrules.jl#L137):
```suggestion
    :NaN, :(  (SpecialFunctions.hankelh2x($ν - 1, $x) - SpecialFunctions.hankelh2x($ν + 1, $x)) / 2 + im * SpecialFunctions.hankelh2x($ν, $x)  )
```

test/runtests.jl
```diff
@@ -18,8 +18,15 @@ end
 
 non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3)]
 
+recently_defined_functions = []
+if !hasmethod(SpecialFunctions.gamma, Tuple{Number, Number})
+    # 2-arg gamma added after SpecialFunctions v0.8.0
+    push!(recently_defined_functions, (:SpecialFunctions, :gamma, 2))
+end
+
```
devmotion, line 26: We should just update the SpecialFunctions compat entry if needed. Might not be necessary anymore now that we only support SpecialFunctions >= 0.10 on the master branch?
```suggestion
```

test/runtests.jl
```diff
@@ -18,8 +18,15 @@ end
...
 for (M, f, arity) in DiffRules.diffrules()
     (M, f, arity) ∈ non_numeric_arg_functions && continue
+    (M, f, arity) ∈ recently_defined_functions && continue
```
devmotion, line 29:
```suggestion
```

devmotion:
I resolved the conflicts and removed the check for `gamma(a, x)`: We should only define a rule for it if it is defined, so IMO the rule should not exist if `SpecialFunctions.gamma(a, x)` does not exist. As discussed in e.g. the PR for `logerfcx` the cleanest approach is to update the compat entry for SpecialFunctions. I updated it to `1.2, 2` since `gamma(a, x)` was added in SpecialFunctions 1.2.0: https://github.com/JuliaMath/SpecialFunctions.jl/compare/v1.1.0...v1.2.0

devmotion:
The ReverseDiff test errors are caused by (correct) NaN values. This is not a problem of this PR but rather of ReverseDiff's tests: https://github.com/JuliaDiff/ReverseDiff.jl/blob/d522508aa6fea16e9716607cdd27d63453bb61e6/test/utils.jl#L9-L13 Maybe this could be fixed (and also the currently skipped functions tested properly) if one uses `isapprox(A, B; nans=true, atol=_atol)` in https://github.com/JuliaDiff/ReverseDiff.jl/blob/d522508aa6fea16e9716607cdd27d63453bb61e6/test/utils.jl#L21?

devmotion:
@mcabbott Are you fine with updating the SpecialFunctions compat entry to "1.2, 2"?

src/rules.jl
```diff
@@ -168,29 +170,34 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
     :NaN, :(  (SpecialFunctions.bessely($ν - 1, $x) - SpecialFunctions.bessely($ν + 1, $x)) / 2  )
 @define_diffrule SpecialFunctions.besselk(ν, x)   =
     :NaN, :( -(SpecialFunctions.besselk($ν - 1, $x) + SpecialFunctions.besselk($ν + 1, $x)) / 2  )
+@define_diffrule SpecialFunctions.besselkx(ν, x)  =
+    :NaN, :( -(SpecialFunctions.besselkx($ν - 1, $x) + SpecialFunctions.besselkx($ν + 1, $x)) / 2 + SpecialFunctions.besselkx($ν, $x)  )
 @define_diffrule SpecialFunctions.hankelh1(ν, x)  =
     :NaN, :(  (SpecialFunctions.hankelh1($ν - 1, $x) - SpecialFunctions.hankelh1($ν + 1, $x)) / 2  )
+@define_diffrule SpecialFunctions.hankelh1x(ν, x) =
+    :NaN, :(  (SpecialFunctions.hankelh1x($ν - 1, $x) - SpecialFunctions.hankelh1x($ν + 1, $x)) / 2 - im * SpecialFunctions.hankelh1x($ν, $x)  )
```
devmotion, line 178: I always assumed that DiffRules only deals with functions with real arguments and real outputs - but `hankelh1x`, `hankelh2x`, and also `hankelh1` and `hankelh2` for which a rule is already defined output complex numbers. So complex numbers are supported officially? It seems based on the derivatives for `conj` and `adjoint` that you added in https://github.com/JuliaDiff/DiffRules.jl/pull/54 you also thought that DiffRules only deals with real numbers @mcabbott?

PR #69, opened 2021-10-21
Add rules for LogExpFunctions

devmotion:
This PR adds the rules for the functions in LogExpFunctions, as defined in https://github.com/JuliaStats/LogExpFunctions.jl/blob/master/src/chainrules.jl.

The motivation for adding them here is that recent versions of SpecialFunctions depend on LogExpFunctions. Hence LogExpFunctions is already an indirect dependency of DiffRules. Moreover, if AD backends such as ForwardDiff, Tracker, and ReverseDiff are fixed (see below) this PR fixes remaining test issues for these backends in DistributionsAD (see https://github.com/TuringLang/DistributionsAD.jl/pull/203).

While this PR is not breaking technically, it seems it breaks almost all downstream dependencies in practice. The reason is that it seems to be common practice to iterate over `DiffRules.diffrules()` and define differentiation rules with `@eval` without checking that the module where the function is defined is available in the current module. I.e., they are all missing a check such as `M in (:Base, :SpecialFunctions, :NaNMath)` to ensure that they don't try to define rules for unsupported packages. I guess this means that we should update the version number to DIffRules 2.0.0 to make sure that we don't break downstream packages. When upgrading to DiffRules 2.0.0, ideally the packages fix this bug.

Additionally, one could provide something like
```julia
macro diffrules() = :(diffrules(__module__))

function diffrules(mod::Module)
    return filter(diffrules()) do (M, _, _)
        return isdefined(mod, M)
    end
end
```
to make it more convenient for packages to load rules for all modules that are loaded in the invoking module.


src/rules.jl
```diff
@@ -232,3 +232,30 @@ end
...
 @define_diffrule NaNMath.min(x, y) = :(ifelse(($y < $x) | (signbit($y) > signbit($x)), ifelse(isnan($y), one($x), zero($x)), ifelse(isnan($x), zero($x), one($x)))),
                                      :(ifelse(($y < $x) | (signbit($y) > signbit($x)), ifelse(isnan($y), zero($y), one($y)), ifelse(isnan($x), one($x), zero($x))))
+
+###################
+# LogExpFunctions #
+###################
+
+# unary
+@define_diffrule LogExpFunctions.xlogx(x) = :(1 + log($x))
+@define_diffrule LogExpFunctions.logistic(x) = :(z = LogExpFunctions.logistic($x); z * (1 - z))
```
devmotion, line 242: Maybe I missed it but it seems there is no way to reuse the result of the primal computation in DiffRules?

devmotion:
Bump :slightly_smiling_face:

andreasnoack:
Is it intended that all the workflow changes are part of this PR? I'd think that they belong in a separate PR. (Btw I don't think we should be using the git based registry).

devmotion:
> Is it intended that all the workflow changes are part of this PR?

Partly. I updated the Julia version in the docs action, updated `docs/make.jl`, and added an action that cleans the docs preview to ensure that the documentation is built correctly and doctests pass.

Additionally, I added the "cancel actions for old commits" feature on purpose since otherwise GH actions would be blocked by the longer running integration tests.

I'll move the TagBot + CompatHelper updates to a separate PR though, they are unrelated.

devmotion:
I moved the CI updates to https://github.com/JuliaDiff/DiffRules.jl/pull/70.

.github/workflows/TagBot.yml
```diff
@@ -1,7 +1,5 @@
 name: TagBot
 on:
-  schedule:
```
devmotion, line 3: @andreasnoack I forgot to remove these lines in #70, seems I copied the updated file incorrectly. `schedule:` is not needed anymore, so removing this trigger will reduce the amount of runs.

All other CI changes were addressed in #70.

andreasnoack (approved):

odow:
> What about updating all the downstream packages before merging this?

This would have been my preference. Arguably the downstream packages are mis-using the API. For packages that are this deep in the dependency tree, we should try to avoid introducing deprecation warnings that users see (we don't want JuMP users seeing deprecation warnings in production or during their tests that they can't do anything about).

Another option would have been to make accessing all the rules opt-in so `diffrules(all = true)`, and the default, `diffrules(all = false)` would have kept the current behavior.

devmotion:
> we should try to avoid introducing deprecation warnings that users see (we don't want JuMP users seeing deprecation warnings in production or during their tests that they can't do anything about).

Users don't see deprecation warnings since they are not enabled by default. And arguably in tests we want downstream packages to see deprecation warnings - they don't have to fix them immediately but they are notified that the API will change in the future and prepare a fix.

odow:
> since they are not enabled by default.

Only in more recent versions of Julia. JuMP still supports the LTS of Julia 1.0.

> in tests we want downstream packages to see deprecation warnings

We want direct dependencies to see them. But if there are lots of transitive dependencies, there are issues. 

DiffRules has 17 direct dependents that might need updating (depending if they called this properly), but 853 indirect dependencies:
image

Most of those are due to ForwardDiff, which _does_ cause the deprecation warning to be thrown:
image

devmotion:
I can understand that you would like that deprecation warnings don't show for users, and apparently also in tests. However, I think the whole point of deprecation warnings is to clearly mark the API as deprecated and notify packages and users about this future change such that they can react accordingly (if they plan to support a future breaking release). In my opinion, deprecation warnings are no errors that have to be fixed immediately since they are no errors but warnings or notifications. It would basically be impossible for packages with many, possibly indirect, dependencies to mark something as deprecated if one had to fix all downstream packages first. And in my opinion this would also invalidate the reason for deprecation warnings. I don't think it can be expected from contributors to fix the use of `diffrules` in all downstream packages such as ForwardDiff, ReverseDiff, or Tracker if there is a clear way to deprecate the current (incorrect) use. (Technically, it would have been even OK to not introduce any deprecation warnings at all since the problem is mainly an incorrect use of `diffrules` in these downstream packages - but this would have been extremely breaking and disruptive for the whole AD ecosystem.)

devmotion:
> Only in more recent versions of Julia. JuMP still supports the LTS of Julia 1.0.

I didn't know it was only introduced in more recent versions. Of course, you can always disable them manually with `--depwarn=no`.

mcabbott:
I think I missed this at the time, but why must this add a dependence on LogExpFunctions?

I mean a test-time dep is fine, and a compat entry in Project.toml is fine (like SpecialFunctions) but why actually load it?

devmotion:
Not all functions are defined in all versions of LogExpFunctions, and it is not possible to check for their existence in any other way than with `idefined(LogExpFunctions, ...)`.

It does not affect dependencies of DiffRule since SpecialFunctions depends on LogExpFunctions and hence it's an indirect dependency anyway. Moreover, it is very lightweight and all its dependencies are very lightweight (only interfaces such as ChainRulesCore, InverseFunctions, and ChangesOfVariables). It was extracted from StatsFuns to make it so lightweight (in particular without Rmath) that it can be added to any package without major impacts on package load times and deoendencies.

mcabbott:
Sorry, yes I only mean changing the bound on SpecialFunctions. The bound on Julia itself is another question, and I agree that would be a very indirect way to constrain anything.

This package did and (I think) still can have *zero* dependencies. That's the best situation, and I think we ought to have a solid reason to move away from it. 

> since SpecialFunctions depends on LogExpFunctions

You mean the reverse? But this we can also solve, there is no real need to depend on either of these.

> and it is not possible to check for their existence in any other way than

But if LogExpFunctions follows semver, then it *is* possible, you just bound the version. Removing them is a breaking change.

mcabbott:
> and it is not possible to check for their existence in any other way than

But if LogExpFunctions follows semver, then surely we can just bound the version. Removing them is a breaking change. This is exactly how SpecialFunctions was handled.

> since SpecialFunctions depends on LogExpFunctions

But this package doesn't depend on SpecialFunctions, except to bound its version. That's the pattern I'm suggesting could be copied. 

This package did and (I think) still can have *zero* dependencies. That's the best situation, and I think we ought to have a solid reason to move away from it. The costs of doing so aren't huge, but they aren't nothing either. It seems brittle to have every little package depend on every other one, e.g. the fairly recent case where some change broke Static.jl and about a thousand downstream packages.

PR #72, opened 2021-11-23
support specialfunctions 2.x

stevengj:
The only breaking change in SpecialFunctions 2.0 [was deleting](https://github.com/JuliaMath/SpecialFunctions.jl/pull/297) the `factorial(x::Real) = gamma(x+1)` function, and since you don't use `factorial` this shouldn't affect you.


devmotion (approved):
Thanks, looks good to me! Tests also used SpecialFunctions 2.0, so this seems fine.

(BTW there was already a CompatHelper PR but it seems there is no (correct) DOCUMENTER_KEY set up on Github and hence it was not allowed to trigger the tests)

PR #73, opened 2021-11-30
Added ldexp to list of diffrules

david-hofmann:
Added a rule for the [function ldexp](https://docs.julialang.org/en/v1/base/math/#Base.Math.ldexp)

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -90,7 +90,7 @@ else
 end
 @define_diffrule Base.hypot(x, y)  = :( $x / hypot($x, $y)                                      ), :(  $y / hypot($x, $y)                                                     )
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
-@define_diffrule Base.ldexp(x, y)  = :( exp2($y) ), :NaN
+@define_diffrule Base.ldexp(x, y)  = :( exp2($y)                                                ), :NaN
 
 @define_diffrule Base.mod(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -floor(z)), NaN)) )
 @define_diffrule Base.rem(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -trunc(z)), NaN)) )
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -90,7 +90,7 @@ else
 end
 @define_diffrule Base.hypot(x, y)  = :( $x / hypot($x, $y)                                      ), :(  $y / hypot($x, $y)                                                     )
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
-@define_diffrule Base.ldexp(x, y)  = :( 2^$y                                                    ), :(log(2) * $x * 2^$y)
+@define_diffrule Base.ldexp(x, y)  = :( exp2($y) ), :NaN
 
 @define_diffrule Base.mod(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -floor(z)), NaN)) )
 @define_diffrule Base.rem(x, y)    = :( first(promote(ifelse(isinteger($x / $y), NaN, 1), NaN)) ), :(  z = $x / $y; first(promote(ifelse(isinteger(z), NaN, -trunc(z)), NaN)) )
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "julia.environmentPath": "/home/david/Projects/0Codes/DiffRules.jl"
+}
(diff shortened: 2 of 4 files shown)


src/rules.jl
```diff
@@ -90,6 +90,7 @@ else
 end
 @define_diffrule Base.hypot(x, y)  = :( $x / hypot($x, $y)                                      ), :(  $y / hypot($x, $y)                                                     )
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
+@define_diffrule Base.ldexp(x, y)  = :( 2^$y                                                    ), :(log(2) * $x * 2^$y)
```
devmotion, line 93: Intuitively, I would say that the derivative of `ldexp` with respect to the second argument is not defined - in ChainRules it would be `NoTangent` (and actually it is: https://github.com/JuliaDiff/ChainRules.jl/blob/ffbaa5fecca8da39f20aeb66cc6e4edf3e0c3f11/src/rulesets/Base/base.jl#L105) and AFAIK in DiffRules this is indicated by `NaN`.

So in my opinion the rule should be defined as
```suggestion
@define_diffrule Base.ldexp(x, y)  = :( exp2($y) ), :NaN
```
Here `exp2` avoids overflow:
```julia
julia> ldexp(1.0, 64)
1.8446744073709552e19

julia> 2^64
0

julia> exp2(64)
1.8446744073709552e19
```

devmotion (changes requested):

src/rules.jl
david-hofmann: Dear @devmotion 
thank you for the comments. The derivative of 2^x is defined: note that 2^x = exp(x\*log(2)), deriving exp(a\*x) w.r.t. x is just a\*exp(a\*x)

Thanks for pointing out the `exp2` feature of avoiding overflows!

Also, I'll note that I am still pretty new and unfamiliar with the workflow in git so please advice if I am missing anything I am supposed to do. Thanks!

devmotion: Sure, one can differentiate `x -> 2^x`. But the crucial point here is that `ldexp` only accepts integers as second argument, and in automatic differentiation, as shown also by the ChainRules definition linked above, integers are considered to be a discrete set and hence non-differentiable. Of course, they could be embedded in the real line but perturbing the input would lead to values that the function can't be called with: `ldexp` only accepts `Integer`s as second argument and errors if it is called with non-integer values (BTW this is also the reason for the test errors.)

This is also explained in the ChainRules documentation: https://juliadiff.org/ChainRulesCore.jl/stable/FAQ.html#faq_abstract_zero

> Also, I'll note that I am still pretty new and unfamiliar with the workflow in git so please advice if I am missing anything I am supposed to do.

In the web interface you can accept my suggestion above by clicking on the "commit suggestion" button below it (if I managed to convince you that it is an improvement :wink:). Otherwise, the main thing left to do is to fix the test errors. This can be done by updating the tests in `test/runtests.jl` and ensuring that the second argument is always an integer, similar to how `rem2pi` is handled: https://github.com/JuliaDiff/DiffRules.jl/blob/68cd4f2416c68bb9ed349bf92e2acc9221591c49/test/runtests.jl#L79

david-hofmann:
Dear @devmotion , I have now also added a special test rule for ldexp. I tested it locally and I think it works as it is supposed to but the test errors still remain. However, as far as I can see it doesn't have anything to do with my code additions or am I missing something? Can I do anything to move forward with the PR? Thank you!

david-hofmann:
Ah, no, nevermind, I see that some are still related to `ldexp`. The ReverseDiff for instance.

david-hofmann:
Okay, I see that at least the issues with the downstream tests of `ForwardDiff` and `ReverseDiff` are related to `ldexp` . It is unclear to me how those tests are run - on my local machine no such tests are run when executing `runtests.jl` . Locally I am passing all the tests (that are being run) successfully. Am I supposed to also alter the tests in ForwardDiff and ReverseDiff to make them compatible with `ldexp`? Thanks for any advice!

devmotion:
> Am I supposed to also alter the tests in ForwardDiff and ReverseDiff to make them compatible with ldexp?

The problem is that these packages define differentiation rules based on the rules in DiffRules automatically and then also test them automatically, without taking into account the types of the arguments of these functions. Of course, this is problematic already for existing rules such as the one for `rem2pi` and hence the rule for `rem2pi` is skipped in the ReverseDiff tests: https://github.com/JuliaDiff/ReverseDiff.jl/search?q=rem2pi (probably would be better to handle it separately but apparently this is not done currently :shrug:). Similarly, in ForwardDiff the rule for `rem2pi` is not tested automatically: https://github.com/JuliaDiff/ForwardDiff.jl/search?q=rem2pi In ForwardDiff it is actually tested separately. Similarly, one would have to skip these automatic tests for `ldexp` and ideally test the rules for it separately.

It would be nice for maintainers of these downstream packages if you could prepare PRs that skip these automatic tests also for `ldexp` - but technically you don't have to: the addition of a new rule is considered non-breaking, and also in these cases it would not break these packages but only their automatic tests. When the rule for `ldexp` is available in DiffRules, then ideally one should add specific tests for `ldexp` to ReverseDiff and ForwardDiff to make sure that their automatic rules are correct.

ChrisRackauckas:
Then let's just merge and leave the rest to downstream?

devmotion:
> Then let's just merge and leave the rest to downstream?

Fine with me. I merged the master branch, this should fix the other test errors.

devmotion (approved):

mcabbott (approved):
Seems OK, breaks ForwardDiff's tests but won't break actual usage.

devmotion:
Thank you @david-hofmann!

PR #74, opened 2021-12-10
AD rule for logerfcx

cossio:
This should fix https://github.com/FluxML/Zygote.jl/issues/1132.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -121,6 +121,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule SpecialFunctions.erfi(x)        = :(  (2 / sqrt(π)) * exp($x * $x)        )
 @define_diffrule SpecialFunctions.erfcx(x)       =
     :(  (2 * $x * SpecialFunctions.erfcx($x)) - (2 / sqrt(π))  )
+@define_diffrule SpecialFunctions.logerfcx(x) =
+    :(  2 * ($x - inv(SpecialFunctions.erfcx($x)) / sqrt(π))  )
 @define_diffrule SpecialFunctions.dawson(x)      =
     :(  1 - (2 * $x * SpecialFunctions.dawson($x))  )
 @define_diffrule SpecialFunctions.digamma(x) =
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -121,8 +121,6 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule SpecialFunctions.erfi(x)        = :(  (2 / sqrt(π)) * exp($x * $x)        )
 @define_diffrule SpecialFunctions.erfcx(x)       =
     :(  (2 * $x * SpecialFunctions.erfcx($x)) - (2 / sqrt(π))  )
-@define_diffrule SpecialFunctions.logerfcx(x) =
-    :(  2 * ($x - inv(SpecialFunctions.erfcx($x)) / sqrt(π))  )
 @define_diffrule SpecialFunctions.dawson(x)      =
     :(  1 - (2 * $x * SpecialFunctions.dawson($x))  )
 @define_diffrule SpecialFunctions.digamma(x) =
@@ -147,6 +145,12 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
     :( -SpecialFunctions.bessely1($x)  )
 @define_diffrule SpecialFunctions.bessely1(x)    =
     :(  (SpecialFunctions.bessely0($x) - SpecialFunctions.bessely(2, $x)) / 2 )
+
+# logerfcx added in SpeciaFunctions 0.10, which requires Julia 1.3
+if VERSION ≥ v"1.3"
+    @define_diffrule SpecialFunctions.logerfcx(x) =
+        :(  2 * ($x - inv(SpecialFunctions.erfcx($x)) / sqrt(π))  )
+end
 
 # TODO:
 #
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -121,6 +121,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule SpecialFunctions.erfi(x)        = :(  (2 / sqrt(π)) * exp($x * $x)        )
 @define_diffrule SpecialFunctions.erfcx(x)       =
     :(  (2 * $x * SpecialFunctions.erfcx($x)) - (2 / sqrt(π))  )
+@define_diffrule SpecialFunctions.logerfcx(x) =
+    :(  2 * ($x - inv(SpecialFunctions.erfcx($x) * sqrt(π)))  )
 @define_diffrule SpecialFunctions.dawson(x)      =
     :(  1 - (2 * $x * SpecialFunctions.dawson($x))  )
 @define_diffrule SpecialFunctions.digamma(x) =
@@ -145,14 +147,6 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
     :( -SpecialFunctions.bessely1($x)  )
 @define_diffrule SpecialFunctions.bessely1(x)    =
     :(  (SpecialFunctions.bessely0($x) - SpecialFunctions.bessely(2, $x)) / 2 )
-
-#= logerfcx added in SpeciaFunctions 0.10, which requires Julia 1.3,
-so we need this check here to make things work in Julia 1.0,
-and in general with older versions of SpecialFunctions =#
-if isdefined(SpecialFunctions, :logerfcx)
-    @define_diffrule SpecialFunctions.logerfcx(x) =
-        :(  2 * ($x - inv(SpecialFunctions.erfcx($x) * sqrt(π)))  )
-end
 
 # TODO:
 #
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -146,8 +146,10 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule SpecialFunctions.bessely1(x)    =
     :(  (SpecialFunctions.bessely0($x) - SpecialFunctions.bessely(2, $x)) / 2 )
 
-# logerfcx added in SpeciaFunctions 0.10, which requires Julia 1.3
-if VERSION ≥ v"1.3"
+#= logerfcx added in SpeciaFunctions 0.10, which requires Julia 1.3,
+so we need this check here to make things work in Julia 1.0,
+and in general with older versions of SpecialFunctions =#
+if isdefined(SpecialFunctions, :logerfcx)
     @define_diffrule SpecialFunctions.logerfcx(x) =
         :(  2 * ($x - inv(SpecialFunctions.erfcx($x) * sqrt(π)))  )
 end
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -149,7 +149,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 # logerfcx added in SpeciaFunctions 0.10, which requires Julia 1.3
 if VERSION ≥ v"1.3"
     @define_diffrule SpecialFunctions.logerfcx(x) =
-        :(  2 * ($x - inv(SpecialFunctions.erfcx($x)) / sqrt(π))  )
+        :(  2 * ($x - inv(SpecialFunctions.erfcx($x) * sqrt(π)))  )
 end
 
 # TODO:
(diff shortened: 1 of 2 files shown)


src/rules.jl
```diff
@@ -150,6 +150,12 @@ end
 @define_diffrule SpecialFunctions.bessely1(x)    =
     :(  (SpecialFunctions.bessely0($x) - SpecialFunctions.bessely(2, $x)) / 2 )
 
+# logerfcx added in SpeciaFunctions 0.10, which requires Julia 1.3
+if VERSION ≥ v"1.3"
```
devmotion, line 154: It would be easier, and in my opinion better, to just check if `logerfcx` is defined (this might require an additional `import SpecialFunctions` in src/DiffRules.jl, don't know if it is loaded currently):
```suggestion
# logerfcx was added in SpeciaFunctions 0.10
if isdefined(SpecialFunctions, :logerfcx)
```

src/rules.jl
```diff
@@ -150,6 +150,12 @@ end
...
+    @define_diffrule SpecialFunctions.logerfcx(x) =
+        :(  2 * ($x - inv(SpecialFunctions.erfcx($x)) / sqrt(π))  )
```
devmotion, line 156: I suggest to use either
```suggestion
        :(  2 * ($x - SpecialFunctions.erfcx(- $x) / sqrt(π))  )
```
or
```suggestion
        :(  2 * ($x - inv(SpecialFunctions.erfcx($x) * sqrt(π)))  )
```

devmotion:
Can also fix the doctest error? I assume either the additional definition or Julia 1.7 leads to a differently ordered set of rules.

cossio:
> Can also fix the doctest error? I assume either the additional definition or Julia 1.7 leads to a differently ordered set of rules.

Regarding this, the current jldoctest relies on a specific order of returned `diffrules`, which is not reliable.
I think it can be fixed by making the order deterministic, https://github.com/JuliaDiff/DiffRules.jl/pull/75.

src/rules.jl
cossio: I think the first suggestion is wrong, did you test it?

Here I just copied the rule defined in ChainRules, which seems to work well.
So, do you have any reason to prefer the second suggestion?

src/rules.jl
cossio: Do we want to `import SpecialFunctions`? Note that will drag dependencies like OpenSpecFun. What if someone just uses ForwardDiff without needing SpecialFunctions?

But I'll follow your advice here.

devmotion: The package already depends on SpecialFunctions.

src/rules.jl
devmotion: It should be correct, it's how the rule is defined in SpecialFunctions. The rules were moved from ChainRules to SpecialFunctions quite a while ago.

devmotion: Ah never mind, I missed that it's exp and not erfcx: https://github.com/JuliaMath/SpecialFunctions.jl/blob/1febdd05902f5b3a3adc9915aba40b776aa3ea24/src/chainrules.jl#L44

devmotion: The second suggestion combines the two divisions, so to me it seemed a bit simpler.

src/rules.jl
cossio: Yes but it's not imported, is it?

src/rules.jl
cossio: Ok I added your second suggestion.

src/rules.jl
devmotion: Yes, it seems in contrast to LogExpFunctions it is not imported currently. However, since it is a dependency it already pulls in OpenSpecfun etc (BTW ForwardDiff depends on SpecialFunctions directly as well) and eg precompilation already includes SpecialFunctions as well.

cossio: Ok I can import SpecialFunctions.
However, maybe the issue is that we shouldn't support Julia 1.0? The new LTS is Julia 1.6. If I replace 1.0 in the ci.yml with 1.6 this issue goes away. What do you think?

cossio: Well nevermind, I'll just do the check you suggest. Can we get this merged as is?

devmotion (approved):
Looks good to me, if tests pass.

src/rules.jl
mcabbott: IIRC it depends on it only to bound the version, without loading it, this is not an accident. Xref #53

mcabbott:
Shouldn't this just change the version bound, rather than check existence of a method? Cc @oxinabox, who had strong opinions the last time this came up.

I think keeping this package at zero dependencies is worth something, although I suspect this is a losing battle and we will all soon depend on a dozen leftpad packages...

src/rules.jl
cossio: Ok what do you suggest then @mcabbott ? I should not load it? How do I check what version of SpecialFunctions is/will be available?

cossio:
I agree increasing the version bound would be best, both of Julia (1.0 -> 1.6, the new LTS), and SpecialFunctions (minimum should be 0.10, so we drop support for 0.8, 0.9).

oxinabox:
> Shouldn't this just change the version bound, rather than check existence of a method? Cc @oxinabox, who had strong opinions the last time this came up.

Checking `isdefined` is best, when that is what you care about.
And doesn't add too much complexity
like in this case.

> I think keeping this package at zero dependencies is worth something

This package has always depended on SpecialFunctions.jl
(well ever since SpecialFunctions.jl was peeled out of Base at least)

cossio:
> This package has always depended on SpecialFunctions.jl

@oxinabox Yes, but it wasn't `import`ed. To check `isdefined` I need to `import` it.
Do we want to import it?

mcabbott:
> has always depended on SpecialFunctions.jl

Well, for some definitions of "depended on ". This package does not load SpecialFunctions, it only has a version bound.

src/rules.jl
devmotion: Sure, it was not an accident - it wasn't necessary to import it. But now it is.

oxinabox:
>   @oxinabox Yes, but it wasn't imported. To check isdefined I need to import it.

Fair point.

Counter, AFAIK every package that depends on DiffRules depends on SpecialFunctions.jl also.
And does load it.
so ::shrug::

src/rules.jl
mcabbott: Well, this is one method, but is it necessary? What's the downside to just changing the version bound?

devmotion: IMO the best solution is to load it. There's no need to update the Julia version bound - it only addresses the problem indirectly and we will want to add more rules that were introduced in SpecialFunctions > 1 for which this hack does not work.

devmotion:
As mentioned in the thread, I don't think updating the Julia compat is a correct approach.

> IMO the best solution is to load it. There's no need to update the Julia version bound - it only addresses the problem indirectly and we will want to add more rules that were introduced in SpecialFunctions > 1 for which this hack does not work.

mcabbott:
A majority, but by no means all, of the packages using this do load SpecialFunctions: https://juliahub.com/ui/Packages/DiffRules/x2ZNk/1.5.0?page=2

Updating the version bound seems like the smallest change, it's just changing the setting on the existing mechanism. There is no reason this won't work for functions added in version > 1, you just move the bound to 1.2, surely? What am I missing?

devmotion:
> There is no reason this won't work for functions added in version > 1, you just move the bound to 1.2, surely? What am I missing?

I was referring to the suggestion of bumping the julua compat to avoid this issue. SpecialFunctions added functions post-1.0 for which we want to define rules at some point for which this hack does not work. Hence I'm against changing the Julia compat.

I would be fine with bumping eg SpecialFunctions compat entry but it's not clear to me yet why `import SpecialFunctions` would be bad. It doesn't affect the dependencies of DiffRules and eg precompilation already includes direct dependencies and hence SpecialFunctions.

mcabbott:
Sorry, yes I only mean changing the bound on SpecialFunctions. The bound on Julia itself is another question, and I agree that would be a very indirect way to constrain anything.

This package did and (I think) still can have *zero* dependencies. That's the best situation, and I think we ought to have a solid reason to move away from it.

cossio:
> Sorry, yes I only mean changing the bound on SpecialFunctions. The bound on Julia itself is another question, and I agree that would be a very indirect way to constrain anything.

If you increase the bound on SpecialFunctions this will also increase the bound on Julia (because SpecialFunctions has a higher lower bound for Julia)

mcabbott:
What bound exactly do you need on SpecialFunctions?

No objection to increasing the bound on Julia as a consequence, the confusion is over increasing the bound on Julia as a *mechanism* to bound SpecialFunctions.

devmotion:
Exactly, it means dropping support for older Julia versions, im- or explicitly. Since we use the same approach (`isdefined`) already for LogExpFunctions, I think the easiest way would be to also use it in this PR and to not touch compat bounds. It's maybe cleaner to discuss and work towards zero dependencies in a separate issue or PR.

mcabbott:
Seems fine to drop < 1.3, no?

#76 reverts this check, back to zero deps.

cossio:
> What bound exactly do you need on SpecialFunctions?

Need SpecialFunctions >= 0.10, which defined logercfx.
In turn, this requires Julia 1.3, https://github.com/JuliaMath/SpecialFunctions.jl/blob/v0.10.0/Project.toml.

cossio:
> Seems fine to drop < 1.3, no?

Ok. But why not jump directly to 1.6, which is the new LTS?

devmotion:
> Seems fine to drop \u003C 1.3, no?

I don't have a strong opinion (I think it's fine) but I think updating julia compat bounds deserves a separate PR to make it more transparent for other contributors and hence easier for them to state their opinion on the matter.

cossio:
> I don't have a strong opinion (I think it's fine) but I think updating julia compat bounds deserves a separate PR to make it more transparent for other contributors and hence easier for them to state their opinion on the matter (and eg to discuss if it should be bumped to 1.6).

Ok. But here we need to make a decision. So at least I can set the bound to 1.3?
I agree that 1.6 can be discussed elsewhere.

However the 1.3 issue is directly relevant to this PR, since Julia 1.3 is a necessary dependence of logerfcx.

devmotion:
As I explained I'm against changing any compat bounds in this PR (not in general, but in this PR).

mcabbott:
IMO 1.6 being LTS is a reason to be more comfortable with dropping 1.0, but there is nothing gained by moving the package's lower bound higher than we have reason to, i.e. higher than 1.3.

I see that CI only runs on 1.3 anyway already, BTW.

Maybe worth making another 1-line PR to bump the lowest Julia version to match? That way anyone watching is more likely to notice.

cossio:
> I see that CI only runs on 1.3 anyway already, BTW.

I took the liberty of adding 1.6 here. But if you prefer we can do that in a separate PR (although I see no reason why).

> Maybe worth making another 1-line PR to bump the lowest Julia version to match? That way anyone watching is more likely to notice.

Sorry, the lowest version where?

mcabbott:
I mean, as David suggest, making an independent PR to change the version bound on Julia in Project.toml. It need do nothing else at all, since CI already does not run on 1.0.

cossio:
> I mean, as David suggest, making an independent PR to change the version bound on Julia in Project.toml. It need do nothing else at all, since CI already does not run on 1.0.

Are you saying that here I should leave Julia 1.0 as the low compat bound, instead of 1.3?

I'm not sure that would work. I mean, if I add the SpecialFunctions >= 0.10 compat bound, and then try to run CI on Julia 1.0, it won't work (because SpecialFunctions 0.10 requirse Julia 1.3).

mcabbott:
No, we're suggesting making another one-line PR which will change the bound.

After that one, this one won't have to change the bound.

cossio:
> No, we're suggesting making another one-line PR which will change the bound.

Ah I see, you suggest to make that PR *before* this one. Ok. If you merge it quickly I can make it right now.

cossio:
@mcabbott @devmotion Would appreciate it if this got merged quickly (if we all agree of course).
https://github.com/JuliaDiff/DiffRules.jl/pull/77

cossio:
All tests passing here. Can we merge this? @mcabbott @devmotion

devmotion (commented):
I had fixed all conflicts already but it seems you force pushed some other changes 😄 In particular, I had already bumped the version number to 1.8.0 - can you update the version number again?

cossio:
> I had fixed all conflicts already but it seems you force pushed some other changes smile In particular, I had already bumped the version number to 1.8.0 - can you update the version number again?

Oh I'm sorry, didn't realize this!
Sure, done.

devmotion (approved):
Looks good to me 👍

cossio:
@mcabbott merge?

cossio:
@mcabbott @devmotion Can we tag a release with this :pray: ?

devmotion:
Done: https://github.com/JuliaRegistries/General/pull/50371

cossio:
Thanks!

PR #75, opened 2021-12-10
Fix doctest failure

cossio:
The order in which DiffRules.diffrules() outputs things is not deterministic (because it depends on the order in which `keys(..)` returns things).

This means this jldoctest can fail:

```julia
julia> first(DiffRules.diffrules())
(:Base, :log2, 1)
```

Here we make it so that the output is deterministic.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -139,8 +139,14 @@ To include all rules, specify `filter_modules = nothing`.
 # Examples
 
 ```jldoctest
-julia> DiffRules.diffrules() |> collect |> sort |> first
-(:Base, :*, 2)
+julia> first(DiffRules.diffrules()) isa Tuple{Symbol,Symbol,Int}
+true
+
+julia> (:Base, :log, 1) in DiffRules.diffrules()
+true
+
+julia> (:Base, :*, 2) in DiffRules.diffrules()
+true
 ```
 
 If you call `diffrules()`, only rules for Base, SpecialFunctions, and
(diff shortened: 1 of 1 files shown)


cossio:
The CI failures here might be related to the changes in Random in Julia 1.7?

src/api.jl
```diff
@@ -139,8 +139,8 @@ To include all rules, specify `filter_modules = nothing`.
 # Examples
 
 ```jldoctest
-julia> first(DiffRules.diffrules())
-(:Base, :log2, 1)
+julia> DiffRules.diffrules() |> collect |> sort |> first
+(:Base, :*, 2)
```
devmotion, line 143: Maybe we should just use more meaningful tests, e.g., something like

```suggestion
julia> first(DiffRules.diffrules()) isa Tuple{Symbol,Symbol,Int}
true

julia> (:Base, :log, 1) in DiffRules.diffrules()
true

julia> (:Base, :*, 2) in DiffRules.diffrules()
true
```
The `collect`, `sort` etc doesn't seem very nice in a documentation and the test is still not completely reliable as it could break if some other rule is added. (A bit unrelated, I'm also not a huge fan of the pipe operator 😛)

cossio: Mm yes I agree. Looks nicer as you suggest. I'll do it.

devmotion (approved):

devmotion:
Thanks for the fix @cossio!

PR #77, opened 2021-12-10
Increase Julia compat 1.3

cossio:
Increase Julia compat bound to 1.3. This is needed for https://github.com/JuliaDiff/DiffRules.jl/pull/74.

Add CI testing on 1.3 (for consistency with compat) and 1.6 (LTS).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
+          - '1.3'
+          - '1.6'
           - '1'
           - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 LogExpFunctions = "0.3.2"
 NaNMath = "0.3"
 SpecialFunctions = "0.8, 0.9, 0.10, 1.0, 2"
-julia = "1"
+julia = "1.3"
 
 [extras]
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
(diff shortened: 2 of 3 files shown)


mcabbott:
Needs version bump, ideally to 1.7, after #76.

cossio:
> Needs version bump, ideally to 1.7, after #76.

Sorry, why do we need a new version when changing a dependency compat bound? This is not implying any change in the API of DiffRules. So why should we increase version here?

mcabbott:
Every release needs a new version. Not a breaking version, so it's more or less free.

And I think it's meant to be a minor version, not patch, in case you ever need to release a bugfix for an older version (although this is unlikely).

cossio:
> Every release needs a new version. Not a breaking version, so it's more or less free.
> 
> And I think it's meant to be a minor version, not patch, in case you ever need to release a bugfix for an older version (although this is unlikely).

Ok. But which PRs are you planning to tag as releases? Maybe we can bundle them together and just do one release?

cossio:
> Every release needs a new version. Not a breaking version, so it's more or less free.
> 
> And I think it's meant to be a minor version, not patch, in case you ever need to release a bugfix for an older version (although this is unlikely).

Ah ok, I think I get it. I bumped the minor version, and rebased. This should be ready to go (as soon as tests pass).

cossio:
The CI on Julia 1.7 is failing on `mod`, by the way. But that should be addressed in a separate PR.
Can we merge this?

cossio:
Regarding the CI test on Julia v.17, see #78

devmotion (commented):
Looks good to me, but let's wait for @mcabbott's approval as well.

Can you also get rid of the branch in https://github.com/JuliaDiff/DiffRules.jl/blob/f46977e53ab7bcf2dd0ae14ae15fcffb635703fe/src/rules.jl#L86-L88? I think it's a good opportunity to remove it :slightly_smiling_face:

cossio:
> Can you also get rid of the branch in

Done

devmotion (approved):
Let's wait for the tests and @mcabbott's approval.

cossio:
Tests pass. @mcabbott do you agree to merge this?

PR #78, opened 2021-12-11
fix mod CI test

cossio:
CI is failing on Julia v1.7. 

The failure is ocurring when testing the rule for `Base.mod`. This function is discontinuous at integer values, so actually I don't understand why tests were passing in previous versions of Julia. In Julia v1.7 the random stream changed and it could be that the issue was just by chance not erroring before, but now it is.

So here I just test `mod` separately, on non-integer arguments.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,7 +13,7 @@ end
 @testset "DiffRules" begin
 @testset "check rules" begin
 
-non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3), (:Base, :mod, 2)]
+non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3)]
 
 for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
     (M, f, arity) ∈ non_numeric_arg_functions && continue
@@ -45,7 +45,11 @@ for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
         derivs = DiffRules.diffrule(M, f, :foo, :bar)
         @eval begin
             let
-                foo, bar = rand(1:10), rand()
+                if "mod" == string($M.$f)
+                    foo, bar = rand() + 13, rand() + 5 # make sure x/y is not integer
+                else
+                    foo, bar = rand(1:10), rand()
+                end
                 dx, dy = $(derivs[1]), $(derivs[2])
                 if !(isnan(dx))
                     @test isapprox(dx, finitediff(z -> $M.$f(z, bar), float(foo)), rtol=0.05)
@@ -89,18 +93,6 @@ for xtype in [:Float64, :BigFloat, :Int64]
                 @test isnan(dy)
             end
         end
-    end
-end
-
-# Treat mod separately because of discontinuities at integers
-derivs = DiffRules.diffrule(:Base, :mod, :x, :y)
-@eval begin
-    let
-        x = randn()
-        y = randn()
-        dx, dy = $(derivs[1]), $(derivs[2])
-        @test isapprox(dx, finitediff(z -> mod(z, y), float(x)), rtol=0.05)
-        @test isapprox(dy, finitediff(z -> mod(x, z), float(y)), rtol=0.05)
     end
 end
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,7 +13,7 @@ end
 @testset "DiffRules" begin
 @testset "check rules" begin
 
-non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3)]
+non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3), (:Base, :mod, 2)]
 
 for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
     (M, f, arity) ∈ non_numeric_arg_functions && continue
@@ -89,6 +89,18 @@ for xtype in [:Float64, :BigFloat, :Int64]
                 @test isnan(dy)
             end
         end
+    end
+end
+
+# Treat mod separately because of discontinuities at integers
+derivs = DiffRules.diffrule(:Base, :mod, :x, :y)
+@eval begin
+    let
+        x = randn()
+        y = randn()
+        dx, dy = $(derivs[1]), $(derivs[2])
+        @test isapprox(dx, finitediff(z -> mod(z, y), float(x)), rtol=0.05)
+        @test isapprox(dy, finitediff(z -> mod(x, z), float(y)), rtol=0.05)
     end
 end
 end


test/runtests.jl
```diff
@@ -13,7 +13,7 @@ end
 @testset "DiffRules" begin
 @testset "check rules" begin
 
-non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3)]
+non_numeric_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ifelse, 3), (:Base, :mod, 2)]
```
devmotion, line 16: This is misleading, the arguments are not non-numeric.

test/runtests.jl
```diff
@@ -91,6 +91,18 @@ for xtype in [:Float64, :BigFloat, :Int64]
         end
     end
 end
+
+# Treat mod separately because of discontinuities at integers
```
devmotion, line 95: I don't think this is the problem, it's not discontinuous at integers in the tests here since we use a floating point number as second argument.

test/runtests.jl
```diff
@@ -91,6 +91,18 @@ for xtype in [:Float64, :BigFloat, :Int64]
...
+derivs = DiffRules.diffrule(:Base, :mod, :x, :y)
+@eval begin
+    let
+        x = randn()
+        y = randn()
+        dx, dy = $(derivs[1]), $(derivs[2])
+        @test isapprox(dx, finitediff(z -> mod(z, y), float(x)), rtol=0.05)
+        @test isapprox(dy, finitediff(z -> mod(x, z), float(y)), rtol=0.05)
+    end
+end
```
devmotion, line 105: IMO it would be preferable to integrate this with the other binary tests above, similar to how different functions and e.g. `mod2pi` are handled in the unary tests.

devmotion (commented):
> This function is discontinuous at integer values, so actually I don't understand why tests were passing in previous versions of Julia.

It's not discontinuous at integers in the tests here since we use a floating point number as second argument.

> In Julia v1.7 the random stream changed and it could be that the issue was just by chance not erroring before, but now it is.

The RNG change is the actual reason, we are just unlucky and end up with random numbers where the very very primitive finite differencing algorithm fails. Maybe we should just use FiniteDifferences.jl instead.

cossio:
> It's not discontinuous at integers in the tests here since we use a floating point number as second argument.

Here is a plot of `mod`.

```julia
using GLMakie
xs = 0:0.001:10; lines(xs, mod.(xs, 2))
```

image

It is discontinuous in the first argument, with a discontinuity every `2` units (or whatever the second argument is).

It so happens that the random (float) value of the first argument used in the (previously failing) test is very near such a point, and the test fails. I thin relying on the ability of the finite difference algorithm near a singularity like this is not the right approach here.

test/runtests.jl
cossio: What the `mod2pi` tests are doing is shifting tested value away from the discontinuity so that the finite diff doesn't blow up. This is essentially what I am doing here by sampling `randn` values, which are not likely to be integers.

test/runtests.jl
cossio: Ok. Then maybe we could have a separate list? Or rename `non_numeric_arg_functions` to `discontinuous_functions`?

devmotion:
If I'm not mistaken the floating point value is the second argument of `mod`. So the example you plotted does not correspond to the tests.

devmotion:
>  I thin relying on the ability of the finite difference algorithm near a singularity like this is not the right approach here.

I assume the problem is that the second argument is too close to 0. A better FD algorithm would already help (and be helpful and more accurate in general) but probably also shifting values to >= 0.5 might help in this specific test.

test/runtests.jl
mcabbott: It's still discontinuous at multiples of the 2nd argument, if I'm thinking right. For finite-differencing to be safe, the 1st argument ought to be safely away from any discontinuity, which isn't guaranteed with two random numbers. I'd pick something like `13+rand(), rand(3:5)`.

cossio:
> Maybe we should just use FiniteDifferences.jl instead.

I have tried with `FiniteDifferences.jl` and it doesn't help. You still get the blowed up differentials.

For reference, the test that's failing is at `mod(2, 0.40000030646823104)`. This hits a discontinuity.

```julia
using GLMakie
ys = 0.3:0.0001:0.5
lines(ys, mod.(2, ys))
```
image

cossio:
> but probably also shifting values to >= 0.5 might help in this specific test

Did you try this? This is a plot of `mod(2, y)` as a function of `y`.

image

Maybe I don't understand but I don't see why the shift would help here.
I mean, the discontinuities get rarer, but there is still a probability of hitting them. And eventually the function is just constant so the test becomes trivial.

Also, that's just the case of `mod(2, y)`. If you try `mod(8, y)` (note that in the tests here, the first argument can reach up to 10), the plot is this:

image

So in this case you would need a larger shift.

test/runtests.jl
devmotion: I agree that it's not guaranteed, hence I suggested using `rand() + 0.5` or `rand() + 1` instead of `rand()` for the second argument. I think this would make it quite unlikely that the first argument `rand(1:10)` is close to a singularity. I assume currently on Julia 1.7 we are unlucky and the second argument is too close to 0.

test/runtests.jl
devmotion: I don't think it has to be handled in a more special way than the different unary functions. A branch in the binary tests or shifting values there should be sufficient.

test/runtests.jl
cossio: Please see the plots I posted below.

devmotion:
> I mean, the discontinuities get rarer, but there is still a probability of hitting them.

It's the main point that it becomes less likely - it means that test failures are less likely.

If necessary, we can also restrict the first argument to smaller values. But intuitively I would assume that it becomes sufficiently unlikely by changing the second argument.

cossio:
`mod(x, y)` has a discontinuity whenever `x / y` is an integer. As @devmotion noted this is more likely the closer `x,y` are to `(0,0)`.

mcabbott:
Seems much easier to pick values where you are *certain* there will be no discontinuities nearby, instead of wondering about probabilities and hoping the seed works out. 

Maybe `x, y = 13+rand(), 5+rand()`?

cossio:
> Seems much easier to pick values where you are _certain_ there will be no discontinuities nearby, instead of wondering about probabilities and hoping the seed works out.
> 
> Maybe `x, y = 13+rand(), 5+rand()`?

Yes, I agree. I just pushed this change :smile: . Can you take a look at the current code?

test/runtests.jl
```diff
@@ -45,7 +45,11 @@ for (M, f, arity) in DiffRules.diffrules(; filter_modules=nothing)
         derivs = DiffRules.diffrule(M, f, :foo, :bar)
         @eval begin
             let
-                foo, bar = rand(1:10), rand()
+                if "mod" == string($M.$f)
+                    foo, bar = rand() + 3, rand() + 7 # make sure x/y is not integer
```
devmotion, line 49: Isn't `mod(foo, bar)` always `foo`? This seems a bit too simple maybe.

devmotion (commented):
I agree, it's probably even better to ensure that it can never cause these problems.

test/runtests.jl
devmotion: Maybe just use @mcabbot's suggestion?

cossio: Hmmm you are right.

cossio:
> Maybe `x, y = 13+rand(), 5+rand()`?

Actually I was using different values, which made the test trivial as @devmotion noticed.

With the values suggested by @mcabbott , we guarantee `x / y` is not an integer:

image

while also guaranteeing that the test is not trivial.

image

devmotion (approved):
Looks good to me, thanks @cossio!

PR #79, opened 2021-12-12
Add missing rules for SpecialFunctions 0.10

devmotion:
This PR is both an alternative and an extension of https://github.com/JuliaDiff/DiffRules.jl/pull/66.

It adds missing rules for functions defined in SpecialFunctions >= 0.10: https://specialfunctions.juliamath.org/v0.10/functions_list

Hence in contrast to #66 it does not add a rule for the incomplete gamma function which is only available in SpecialFunctions >= 1.2 and also does not require to update the SpecialFunctions compat entry.

I went through the list of functions in SpecialFunctions 0.10 and added rules that were missing for functions that are differentiable with respect to at least one argument and return scalar values (some functions return tuples but AFAIK these are not supported by DiffRules). Some functions are not holomorphic (e.g. `besselix`) and hence derivatives are only correct for real-valued inputs. It seems DiffRules does not support non-holomorphic functions and hence I only copied the rules for real inputs but not the ones for complex inputs from e.g. https://github.com/JuliaMath/SpecialFunctions.jl/blob/1febdd05902f5b3a3adc9915aba40b776aa3ea24/src/chainrules.jl#L198-L223; I also added some explanations to the code.

To make sure that the rules are defined correctly I added FiniteDifferences as a test dependency and decreased relative and absolute tolerances to 1e-9 (not possible with the current basic finite differencing algorithm).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,8 +15,9 @@ SpecialFunctions = "0.10, 1.0, 2"
 julia = "1.3"
 
 [extras]
+FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Random"]
+test = ["FiniteDifferences", "Test", "Random"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DiffRules"
 uuid = "b552c78f-8df3-52c6-915a-8e097449b14b"
-version = "1.9.0"
+version = "1.10.0"
 
 [deps]
 LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -189,6 +189,9 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 # binary #
 #--------#
 
+@define_diffrule SpecialFunctions.erf(x, y) =
+    :(  -2 / sqrt(π) * exp(-x^2)  ), :(  2 / sqrt(π) * exp(-y^2)  )
+
 # derivatives with respect to the order `ν` exist but are not implemented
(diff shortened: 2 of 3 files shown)


devmotion:
ForwardDiff tests should be fixed by https://github.com/JuliaDiff/ForwardDiff.jl/pull/568.

devmotion:
Bump :slightly_smiling_face: 

Recently multiple users reported problems on Slack with differentiating some of the special functions added in this PR, e.g., `logerfc` or `besselix`.

ReverseDiff and ForwardDiff define and test all rules unconditionally - even for functions with complex outputs. Therefore tests fail.

mcabbott (approved):
This looks fine to me. I haven't checked the maths, but happy to believe you have, and that tests will catch most things. 

I think ForwardDiff only defines rules for `f(::Dual)` and only allows `Dual(::Real)`, hence `Complex{<:Dual}` not the reverse.

(Seems fine to increase bound on SpecialFuncitons too, but that can be later.)

devmotion:
I'll prepare PRs to ForwardDiff and ReverseDiff that fix the test errors before merging this PR.

devmotion:
I made a PR that fixes ForwardDiff more generally: https://github.com/JuliaDiff/ForwardDiff.jl/pull/577

andreasnoack:
Let's try to rerun CI here

andreasnoack:
What should we do with the ReverseDiff tests? Will you prepare a similar PR?

devmotion:
I'm not sure if ReverseDiff can be updated similarly without major changes. IIRC the type signatures were too restrictive when I checked it a while ago. Maybe the easiest way forward would be to just disable the failing rules for now in ReverseDiff.

Moelf (approved):

devmotion:
I had a look at the ReverseDiff test errors, it seems https://github.com/JuliaDiff/ReverseDiff.jl/pull/209 would fix all of the issues related to this PR here.

devmotion:
With the latest release of ReverseDiff tests pass. The test failures of EAGO seem unrelated.

devmotion:
Since you commented above @andreasnoack: Do you think the PR can be merged now?

andreasnoack (approved):

PR #80, opened 2021-12-23
Use FiniteDifferences in tests

devmotion:
I extracted this PR from https://github.com/JuliaDiff/DiffRules.jl/pull/79, I guess this makes it easier to review.

This PR replaces the custom and very basic forward finite differencing method in the tests with FiniteDifferences (e.g. also used by ChainRulesTestUtils). This allows us to lower the tolerances quite significantly which hopefully uncovers incorrect implementations more reliably.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,8 +15,9 @@ SpecialFunctions = "0.10, 1.0, 2"
 julia = "1.3"
 
 [extras]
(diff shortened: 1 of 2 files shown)


torfjelde:
Seems like everyting passes except Nabla.jl (which I presume is just because Nabla isn't comatible with Julia 1.7?).

devmotion:
Fixed by https://github.com/JuliaDiff/DiffRules.jl/pull/55.

PR #82, opened 2022-04-27
Add diffrules for `log1pmx` and `logmxp1`

simsurace:
There may be more optimal orders of operations for these derivatives.
Also, could someone assist me in what would be mandatory tests for these additions?

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -244,6 +244,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule LogExpFunctions.log1mexp(x) = :(-exp($x - LogExpFunctions.log1mexp($x)))
 @define_diffrule LogExpFunctions.log2mexp(x) = :(-exp($x - LogExpFunctions.log2mexp($x)))
 @define_diffrule LogExpFunctions.logexpm1(x) = :(exp($x - LogExpFunctions.logexpm1($x)))
+@define_diffrule LogExpFunctions.log1pmx(x) = :(-$x / (1 + $x))
+@define_diffrule LogExpFunctions.logmxp1(x) = :(inv($x) - 1)
 
 # binary
 @define_diffrule LogExpFunctions.xlogy(x, y) = :(log($y)), :($x / $y)


simsurace:
Looks like https://github.com/JuliaStats/LogExpFunctions.jl/pull/45 is actually a prerequisite.

devmotion:
It's already tested automatically but maybe it's helpful to adjust the domain of inputs (both for the primal and the finite diff comparisons), similar to some other functions in runtests.jl. I don't remember the default domain by heart, so maybe it's not needed at all.

Ironically, tests fail because the functions are not defined for `Float32` (yet). So for the time being, probably we should skip `Float32` tests for these inputs (or check that they throw an error, to be notified when this changes).

src/rules.jl
```diff
@@ -244,6 +244,8 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule LogExpFunctions.log1mexp(x) = :(-exp($x - LogExpFunctions.log1mexp($x)))
 @define_diffrule LogExpFunctions.log2mexp(x) = :(-exp($x - LogExpFunctions.log2mexp($x)))
 @define_diffrule LogExpFunctions.logexpm1(x) = :(exp($x - LogExpFunctions.logexpm1($x)))
+@define_diffrule LogExpFunctions.log1pmx(x) = :(-$x / (1 + $x))
+@define_diffrule LogExpFunctions.logmxp1(x) = :(inv($x) - 1)
```
devmotion, line 248: Just curious, any particular reason to use this instead of
```suggestion
@define_diffrule LogExpFunctions.logmxp1(x) = :((1 - $x) / $x)
```
which would be more similar to the line above?

test/runtests.jl
```diff
@@ -40,7 +40,11 @@ non_diffeable_arg_functions = [(:Base, :rem2pi, 2), (:Base, :ldexp, 2), (:Base,
                     # We're happy with types with the correct promotion behavior, e.g.
                     # it's fine to return `1` as a derivative despite input being `Float64`.
                     @test promote_type(typeof($deriv), $T) === $T
-                    @test $deriv ≈ finitediff($M.$f, goo) rtol=1e-2 atol=1e-3
+                    if $(f in (:log1pmx, :logmxp1)) && $T == Float32
+                        @test_throws MethodError finitediff($M.$f, goo)
+                    else
+                        @test $deriv ≈ finitediff($M.$f, goo) rtol=1e-2 atol=1e-3
```
devmotion, line 46: Seems good. Can you add `logmxp1` to the `rand($T) + $T(0.5)` branch above? That should make finite differencing even less likely to error.

devmotion (commented):
Thanks, looks good! I just have one suggestion to make finite differencing even less likely to fail.

src/rules.jl
simsurace: I don't know if any of them would be preferable. Are there floating-point concerns?

test/runtests.jl
simsurace: Will do!

src/rules.jl
devmotion: Honestly, I don't know. I was just surprised by the two different approaches in this PR and became curious. I used `(1 - x) / x` for the ChainRules definitions in LogExpFunctions now. In general, sometimes a reason to avoid `inv(x)` would be undesired promotions but in this case here it doesn't matter (as far as I can see).

simsurace: I will use the same as in ChainRules then for consistency

devmotion (approved):
Looks good to me, thank you!

I'll merge and release once tests pass (and feel free to ping me if I forget it :slightly_smiling_face:).

devmotion:
The MTK test errors are unrelated.

PR #85, opened 2022-07-07
Add `iszero(x)` branches to `xlogy` and `xlog1py`

simsurace:
It is unclear to me how the tests work here. I did not find the tests of corner cases.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -259,4 +259,4 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 
 @define_diffrule LogExpFunctions.xlog1py(x, y) = 
     :(log1p($y)),
-    :(yp1 = 1 + $y; z = $x / yp1; iszero($x) && !isnan($y) ? zero(z) : iszero(yp1) ? oftype(z, NaN) : z)
+    :(z = $x / (1 + $y); iszero($x) && !isnan($y) ? zero(z) : z)
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -248,11 +248,15 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule LogExpFunctions.logmxp1(x) = :((1 - $x) / $x)
 
 # binary
-@define_diffrule LogExpFunctions.xlogy(x, y) = :(log($y)), :($x / $y)
+@define_diffrule LogExpFunctions.xlogy(x, y) = 
+    :(log($y)),
+    :(z = $x / $y; iszero($x) && !isnan($y) ? zero(z) : z)
 @define_diffrule LogExpFunctions.logaddexp(x, y) =
     :(exp($x - LogExpFunctions.logaddexp($x, $y))), :(exp($y - LogExpFunctions.logaddexp($x, $y)))
 @define_diffrule LogExpFunctions.logsubexp(x, y) =
     :(z = LogExpFunctions.logsubexp($x, $y); $x > $y ? exp($x - z) : -exp($x - z)),
     :(z = LogExpFunctions.logsubexp($x, $y); $x > $y ? -exp($y - z) : exp($y - z))
 
-@define_diffrule LogExpFunctions.xlog1py(x, y) = :(log1p($y)), :($x / (1 + $y))
+@define_diffrule LogExpFunctions.xlog1py(x, y) = 
+    :(log1p($y)),
+    :(z = $x / (1 + $y); iszero($x) && !isnan($y) ? zero(z) : z)
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -250,7 +250,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 # binary
 @define_diffrule LogExpFunctions.xlogy(x, y) = 
     :(log($y)),
-    :(z = $x / $y; iszero($x) && !isnan($y) ? zero(z) : z)
+    :(z = $x / $y; iszero($x) && !isnan($y) ? zero(z) : iszero($y) ? oftype(z, NaN) : z)
 @define_diffrule LogExpFunctions.logaddexp(x, y) =
     :(exp($x - LogExpFunctions.logaddexp($x, $y))), :(exp($y - LogExpFunctions.logaddexp($x, $y)))
 @define_diffrule LogExpFunctions.logsubexp(x, y) =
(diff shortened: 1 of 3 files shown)


src/rules.jl
```diff
@@ -248,11 +248,15 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule LogExpFunctions.logmxp1(x) = :((1 - $x) / $x)
 
 # binary
-@define_diffrule LogExpFunctions.xlogy(x, y) = :(log($y)), :($x / $y)
+@define_diffrule LogExpFunctions.xlogy(x, y) = 
+    :(log($y)),
+    :(z = $x / $y; iszero($x) && !isnan($y) ? zero(z) : iszero($y) ? oftype(z, NaN) : z)
```
devmotion, line 253: Maybe we rather want

```suggestion
    :(z = $x / $y; iszero($x) && !isnan($y) ? zero(z) : z)
```

here. See https://github.com/JuliaStats/LogExpFunctions.jl/pull/54#discussion_r916492896. An additional reason is that otherwise `z = NaN` might be changed to a different `NaN`.

src/rules.jl
```diff
@@ -248,11 +248,15 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
...
 @define_diffrule LogExpFunctions.logaddexp(x, y) =
     :(exp($x - LogExpFunctions.logaddexp($x, $y))), :(exp($y - LogExpFunctions.logaddexp($x, $y)))
 @define_diffrule LogExpFunctions.logsubexp(x, y) =
     :(z = LogExpFunctions.logsubexp($x, $y); $x > $y ? exp($x - z) : -exp($x - z)),
     :(z = LogExpFunctions.logsubexp($x, $y); $x > $y ? -exp($y - z) : exp($y - z))
 
-@define_diffrule LogExpFunctions.xlog1py(x, y) = :(log1p($y)), :($x / (1 + $y))
+@define_diffrule LogExpFunctions.xlog1py(x, y) = 
+    :(log1p($y)),
+    :(yp1 = 1 + $y; z = $x / yp1; iszero($x) && !isnan($y) ? zero(z) : iszero(yp1) ? oftype(z, NaN) : z)
```
devmotion, line 262: Same here:

```suggestion
    :(z = $x / (1 + $y); iszero($x) && !isnan($y) ? zero(z) : z)
```

src/rules.jl
simsurace: This is what I had before. The NaN conventions are misterious to me. I think the -Inf is more informative, but this will make it impossible to match with finite differencing (I think).

simsurace:
Is this fine to be merged?

src/rules.jl
```diff
@@ -248,11 +248,15 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule LogExpFunctions.logmxp1(x) = :((1 - $x) / $x)
 
 # binary
-@define_diffrule LogExpFunctions.xlogy(x, y) = :(log($y)), :($x / $y)
+@define_diffrule LogExpFunctions.xlogy(x, y) = 
+    :(log($y)),
+    :(z = $x / $y; iszero($x) && !isnan($y) ? zero(z) : z)
```
devmotion, line 253: This is not consistent with the rules in https://github.com/JuliaStats/LogExpFunctions.jl/pull/54, so before merging any of these PRs we should converge.

https://github.com/JuliaStats/LogExpFunctions.jl/pull/54 currently implements:
```suggestion
    :(logy = log($y); iszero($y) ? oftype(logy, NaN) : logy),
    :(w = $x / $y; iszero($x) && !isnan($y) ? zero(w) : (iszero(y) ? oftype(w, NaN) : w)
```

Similarly, the definition below is different from the one in LogExpFunctions.

simsurace: Based on your objection in https://github.com/JuliaStats/LogExpFunctions.jl/pull/54, it seems that this PR should stay as it is and that PR should be changed, or am I misinterpreting what you wrote? 

There is still an open conversation regarding https://github.com/JuliaStats/LogExpFunctions.jl/pull/54.

Your suggestion above would take us back to commit https://github.com/JuliaDiff/DiffRules.jl/pull/85/commits/fd1902da9f7e1488da73eb6f690515ffb8d55d30, at least for line 253. Is that what you would like to do?

devmotion: I don't have made up my mind regarding `NaN` - usually they are annoying but maybe that's the right thing to do here. I think the main problem with https://github.com/JuliaStats/LogExpFunctions.jl/pull/54 currently is that tests don't pass?

simsurace: Yes the tests don't pass, presumably due to the NaN treatment not matching finite differencing. I need to dig in. But I think the question of which branches to keep is kind of independent, as the tests will have to take this into account.

I guess we may choose the versions with the lowest number of branches for now, and add them later as needed?

devmotion: Sure, that might be a good guideline.

simsurace:
Then I think we can merge and I'll change https://github.com/JuliaStats/LogExpFunctions.jl/pull/54 to follow the same guidelines.

devmotion (commented):
Looks OK. We want to use the same rules in LogExpFunctions. Unfortunately, currently these rules cause test failures: https://github.com/JuliaStats/LogExpFunctions.jl/pull/57 So let us wait with this PR until we have solved these issues in LogExpFunctions, in case we have to handle some other special case here.

simsurace:
This should be fine to be merged now as well.

devmotion (commented):
Test failures are unrelated, see https://github.com/FluxML/Tracker.jl/issues/125. Similar issues seem to exist for ReverseDiff on Julia 1.8.

Can we add some tests of the new branches though?

simsurace:
Sure. I just don't know where to put them, since there currently do not seem to be _any_ tests for `xlogy` and `xlog1py`. Or where should I look for them?

devmotion:
> Sure. I just don't know where to put them, since there currently do not seem to be any tests for xlogy and xlog1py.

All rules apart from the ones excluded in https://github.com/JuliaDiff/DiffRules.jl/blob/2c29d7d59c09f1ca078192fa24eef8d7fba7738e/test/runtests.jl#L17 are tested in https://github.com/JuliaDiff/DiffRules.jl/blob/2c29d7d59c09f1ca078192fa24eef8d7fba7738e/test/runtests.jl#L19-L110. 

Tests of such special cases might be best to add below https://github.com/JuliaDiff/DiffRules.jl/blob/2c29d7d59c09f1ca078192fa24eef8d7fba7738e/test/runtests.jl#L142, i.e., below the special tests for `rem2pi` and `ldexp`.

Issue #88, opened 2022-11-02
Type instability in ldexp with Float32 arguments


ptiede:
Hi,

This is a cross-post with https://github.com/JuliaDiff/ForwardDiff.jl/issues/604.
The rule for `ldexp(x,y)` always returns a Float64 regardless of the type of the first argument `x`.
This is because the rule is given by

```julia
@define_diffrule Base.ldexp(x, y)  = :( exp2($y) ), :NaN
```
and since `y` is an `Int` `exp2` by default return a Float64. @mcabbott pointed out that if we change the rule to 

```julia
@define_diffrule Base.ldexp(x, y)  = :(  oftype(float($x), exp2($y)  ), :NaN
```
then we should maintain the Float32 type throughout the computation.

ptiede:
Closed because of #89

PR #89, opened 2022-11-03
Preserve input types for various rules

ptiede:
This pull request ensures that the input type is preserved for various rules. 
Previously there were potentially a few places where 64 bit NaN's would always be produced
regardless of the input. To fix this I replaced any instance of `:NaN` with `oftype($x, NaN)`.

Additionally, this pull-request fixes the issue with the `ldexp` rule from #88 which is similar in nature.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -89,7 +89,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 
 @define_diffrule Base.mod(x, y)    = :( z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), one(float(z))) ), :(  z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), -floor(float(z))) )
 @define_diffrule Base.rem(x, y)    = :( z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), one(float(z))) ), :(  z = $x / $y; ifelse(isinteger(z), oftype(float(z), NaN), -trunc(float(z))) )
-@define_diffrule Base.rem2pi(x, r) = :( 1                                                       ), :(oftype(float($x), NaN))
+@define_diffrule Base.rem2pi(x, r) = :( 1                                                       ), :NaN
 @define_diffrule Base.max(x, y)    = :( $x > $y ? one($x) : zero($x)                            ), :( $x > $y ? zero($y) : one($y)                                            )
 @define_diffrule Base.min(x, y)    = :( $x > $y ? zero($x) : one($x)                            ), :( $x > $y ? one($y) : zero($y)                                            )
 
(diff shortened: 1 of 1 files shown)


devmotion:
> To fix this I replaced any instance of :NaN with oftype($x, NaN).

`:NaN` is the DiffRules way to say that some derivative does not exist or is not implemented. You never want to use these "derivatives" anyway, every time you do you are already screwed.

Generally, `oftype($x, NaN)` seems like the wrong approach as `x` might not even be a floating point number. Just using `float($x)` instead seems also wrong in the two-argument functions since it might cause undesired promotions if `x` is not a floating point number (imagine eg the other argument being of type `Float32`).

ptiede:
Ok I reverted the `:NaN` for the two argument functions.

However for the `ldexp` I think the point remains. In that case you do need the `oftype(float($x), exp2($y))` to prevent a spurious promotion for the derivative with respect to `x`.

src/rules.jl
```diff
@@ -85,7 +85,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.atan(x, y)    = :( $y / ($x^2 + $y^2)                                 ), :( -$x / ($x^2 + $y^2)                                                     )
 @define_diffrule Base.hypot(x, y)  = :( $x / hypot($x, $y)                                      ), :(  $y / hypot($x, $y)                                                     )
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
-@define_diffrule Base.ldexp(x, y)  = :( exp2($y)                                                ), :NaN
+@define_diffrule Base.ldexp(x, y)  = :( oftype(float($x), exp2($y))                                                ), :(oftype(float($x), NaN))
```
devmotion, line 88: Could we even just use
```suggestion
@define_diffrule Base.ldexp(x, y)  = :( oftype($x, exp2($y) ), :NaN
```
? At least it seems the definitions in Base already assume that `x` is a floating point number:
```julia
julia> methods(ldexp)
# 7 methods for generic function "ldexp":
[1] ldexp(x::Float16, q::Integer) in Base.Math at math.jl:826
[2] ldexp(x::T, e::Integer) where T<:Union{Float16, Float32, Float64} in Base.Math at math.jl:783
[3] ldexp(x::BigFloat, n::Int64) in Base.MPFR at mpfr.jl:648
[4] ldexp(x::BigFloat, n::Union{Int16, Int32, Int64, Int8}) in Base.MPFR at mpfr.jl:658
[5] ldexp(x::BigFloat, n::UInt64) in Base.MPFR at mpfr.jl:653
[6] ldexp(x::BigFloat, n::Union{UInt16, UInt32, UInt64, UInt8}) in Base.MPFR at mpfr.jl:659
[7] ldexp(x::BigFloat, n::Integer) in Base.MPFR at mpfr.jl:660
```
But maybe
```suggestion
@define_diffrule Base.ldexp(x, y)  = :( oftype(float($x), exp2($y) ), :NaN
```
is safer - although maybe even safer would be something like `oftype(ldexp($x, $y), ...` (or something similar without evaluating the primal) in case the arguments are promoted in some way in some other, non-Base definitions.

ptiede: I tend to agree for the first option. I would be worried about including the primal in the calculation since that is a much more expensive operation than exp2(y) which is just some bit shuffling since `y` is a integer.

src/rules.jl
```diff
@@ -85,7 +85,7 @@ _abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 @define_diffrule Base.atan(x, y)    = :( $y / ($x^2 + $y^2)                                 ), :( -$x / ($x^2 + $y^2)                                                     )
 @define_diffrule Base.hypot(x, y)  = :( $x / hypot($x, $y)                                      ), :(  $y / hypot($x, $y)                                                     )
 @define_diffrule Base.log(b, x)    = :( log($x) * inv(-log($b)^2 * $b)                          ), :( inv($x) / log($b)                                                       )
-@define_diffrule Base.ldexp(x, y)  = :( exp2($y)                                                ), :NaN
+@define_diffrule Base.ldexp(x, y)  = :( oftype($x, exp2($y))                                                ), :(oftype(float($x), NaN))
```
devmotion, line 88: I don't think `NaN` has to be converted here?
```suggestion
@define_diffrule Base.ldexp(x, y)  = :( oftype($x, exp2($y))                                                ), :NaN
```

ptiede: Sorry I missed that one!

devmotion (approved):
Looks good to me.

PR #91, opened 2023-02-20
Support IrrationalConstants 0.2

devmotion:
IrrationalConstants 0.2 changes the type of the irrational constants to fix a type piracy issue: https://github.com/JuliaMath/IrrationalConstants.jl/pull/19 This change is breaking but should not cause any problems in DiffRules.


devmotion:
Test failures are unrelated and exist also on the master branch.

PR #99, opened 2023-06-02
Derivative of exponential integral

ashiklom:
General form:

$$ \frac{d}{dx} E_\nu(x) = -E_{\nu - 1}(x) $$

Note that:

$$ E_0(x) = \frac{e^{-x}}{x} $$

Therefore, derivative of `SpecialFunctions.expint(x)` ($` E_1(x) `$) is:

$$ \frac{d}{dx} E_1(x) = -E_0(x) = -\frac{e^{-x}}{x} $$

Source: [Wikipedia](https://en.wikipedia.org/wiki/Exponential_integral#Derivatives)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 IrrationalConstants = "0.1.1, 0.2"
 LogExpFunctions = "0.3.2"
 NaNMath = "0.3, 1"
-SpecialFunctions = "0.10, 1.0, 2"
+SpecialFunctions = "1.1, 2"
 julia = "1.3"
 
 [extras]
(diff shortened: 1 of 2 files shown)


devmotion:
Looks mainly good (it also matches the rules we added to SpecialFunctions, e.g., https://github.com/JuliaMath/SpecialFunctions.jl/blob/ae35d10713a470b852e53e0c79f20252b5572fa9/ext/SpecialFunctionsChainRulesCoreExt.jl#L183). I am pretty sure though that it requires us to update the SpecialFunctions compat entry since IIRC these functions require a relatively recent release of SpecialFunctions. Can you check this?

ashiklom:
Good catch! Looks like this starts to work (or at least, all the tests pass) starting in SpecialFunctions v1.1. Bumped compatibility accordingly.

devmotion (approved):

PR #100, opened 2023-06-03
Revert "Update abs diff rule to 0 at non-differentiable point"

devmotion:
Reverts JuliaDiff/DiffRules.jl#98, due to the problem explained in https://github.com/JuliaDiff/DiffRules.jl/pull/98#issuecomment-1574420052.

Maybe this issue explains why the "derivative" at 0 was defined as 1 originally (first in ForwardDiff, and then in DiffRules when thw rule was moved here) - but unfortunately neither DiffRules nor ForwardDiff tests it.

In principle, it seems that one could argue for any subderivative, ie., any number in [-1, 1], as the "derivative" at 0. The `abs` function is also discussed in the ChainRules docs, where it is arfued for taking the subderivative 0 and against -1 or 1. However, yhe example in https://github.com/JuliaDiff/DiffRules.jl/pull/98#issuecomment-1574420052 provides an argument for taking 1.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -70,7 +70,7 @@
 
 # We provide this hook for special number types like `Interval`
 # that need their own special definition of `abs`.
-_abs_deriv(x) = sign(x)
+_abs_deriv(x) = signbit(x) ? -one(x) : one(x)
 
 # binary #
 #--------#


devmotion:
Pinging @agerlach and @oxinabox since they were involved in #98.

agerlach:
@devmotion My apologies for the delay.

Given the downstream issues caused, I see no reason why not to revert #98.

oxinabox:
this seems fine, would be good to fix the comment though.
Since the comment there talked about working with Interval types.
And the changed version worked better with interval types.

The hessian coming out at zero is just a special case of the derivative being computered as zero for
```
function f(x)
     if x==1.0
        1.0
     else
        x
     end
end

f'(1.0) == 0
```
right?

To which this is not a satisfying solution but it is better than breaking actual code, for conceptual improvements

devmotion:
I guess the comment is correct but possibly should be clarified: the `_abs_deriv` function was intended as a hook for eg interval packages - they were supposed to implement it for interval types.

devmotion:
I agree, it's not very satisfying that we run into problems when defining the derivative at 0 as 0. Maybe it's generally safer to take the left or right limit, as long as such special branches lead to zero derivatives in e.g. ForwardDiff?

devmotion:
BTW I checked and the problem shows up also with ForwardDiff#master.

